Issue #12, opened 2020-01-31
most parts of Cluster should be covered by unit tests


adietish:
the class Cluster is a crucial component and thus all major parts should get covered by unit tests

adietish:
fixed by PR #10

PR #10, opened 2020-01-24
support switchable "current namespace" (#9)

adietish:
fixes #9

PR #13, opened 2020-01-31
added tests for cluster, corrected API in cluster to fullfil tests (#12)

adietish:
depends on #10 which needs to be merged first


adietish:
closing so that it can be replaced by draft PR

Issue #9, opened 2020-01-23
support switchable "current namespace"


adietish:
The new interaction concept - matching vscode-kubernetes but also oc, kubectl - builds upon the notion of a "current namespace". The resources that the tree displays belong to this "current namespace".
A user can switch the "current namespace" by double clicking on a namespace listed in the tree, which will then cause all resource categories to refresh and watches to be reinitialized

PR #16, opened 2020-01-31
impl'd action to set current namespace

### LINKED ISSUE #15: allow switching "current namespace" in UI
We should allow the user to switch the "current namespace" that all resources are displayed for. In VSCode this is available via "Use Namespace" entry in the context menu
image

adietish:
depends on #9 
fixes #15


adietish:
@sbouchet @jeffmaury could you please review/test this PR?

Issue #4, opened 2020-01-10
sort namespace resources


adietish:
Currently pods that are listed in the tree as children for a namespace are listed in no specific order. Worse even, if you refresh the namespace, their ordering is all different. Same happens if updates happen (ex. you remove or create a pod).
We should ensure a specific (ex. alphabetical oder) of child resources for a namespace

adietish:
fixed by #17. Closing.

PR #17, opened 2020-02-03
#4 sort namespaces and pods according to their labels

bdshadow:
Signed-off-by: Dmitrii Bocharov <bdshadow@gmail.com>

also sorted the namespaces accordingly.

The sorting also can be done by passing a Comparator to `StructureTreeModel` in `KubernetesToolWindowFactory`, but it will be applied to the whole tree, as far as I understand. So I thought that sorting it separately is better. Correct me, please, if I'm wrong, I'll fix it.


adietish:
@bdshadow wow, what a nice suprise. Very glad to hear from you again :)

bdshadow:
@adietish  thank you :)

PR #22, opened 2020-03-02
catch exceptions when connecting to cluster & notify (#20)

### LINKED ISSUE #20: Display error if current cluster is not reachable
Currently if the current cluster isn't reachable nothing is displayed. We should display the cluster node and an error telling the user that it could not be reached:

image

adietish:
fixes #20

### DIFF
--- a/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/Cluster.kt
+++ b/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/Cluster.kt
@@ -10,11 +10,14 @@
  ******************************************************************************/
 package org.jboss.tools.intellij.kubernetes.model
 
+import com.intellij.openapi.diagnostic.logger
 import io.fabric8.kubernetes.api.model.HasMetadata
 import io.fabric8.kubernetes.api.model.Namespace
 import io.fabric8.kubernetes.client.ConfigBuilder
 import io.fabric8.kubernetes.client.DefaultKubernetesClient
+import io.fabric8.kubernetes.client.KubernetesClientException
 import io.fabric8.kubernetes.client.NamespacedKubernetesClient
+
 
 interface ICluster {
     val client: NamespacedKubernetesClient
@@ -24,7 +27,7 @@ interface ICluster {
     fun getAllNamespaces(): List<Namespace>
     fun setCurrentNamespace(namespace: Namespace)
     fun getCurrentNamespace(): Namespace?
-    fun getNamespace(name: String): Namespace?
+    fun getNamespace(name: String?): Namespace?
     fun getNamespaceProvider(name: String?): NamespaceProvider?
     fun getNamespaceProvider(resource: HasMetadata): NamespaceProvider?
     fun getNamespaceProvider(namespace: Namespace): NamespaceProvider?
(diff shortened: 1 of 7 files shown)


adietish:
@jeffmaury, @sbouchet please review

steps: 
1. ASSERT: have a minikube cluster running
1. ASSERT: make sure minikube cluster is current-context (via ```kubectl config use-context <name>```)
1. EXEC: shut down minikube
1. EXEC: launch plugin

Result: 
Tree shows up with error as child to **Namespaces** & **Workloads/Pods**. 
image
You should also get an error notification and a entry in the "Event Log"
image

adietish:
@jeffmaury , @sbouchet: What do you think of the notifications? Worth it? I also tend to think that the errors in the tree should be simplified text-wise (to be more like the popups). Thoughts?

adietish:
improved the errors that are shown in the tree:
image
@jeffmaury, @sbouchet: Thoughts?

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/Cluster.kt
```diff
@@ -120,7 +132,11 @@ open class Cluster(private val modelChange: IModelChangeObservable) : ICluster {
     }
 
     private fun loadAllNamespaces(): Sequence<Namespace> {
-        return  client.namespaces().list().items.asSequence()
+        try {
+            return client.namespaces().list().items.asSequence()
+        } catch (e: KubernetesClientException) {
+            throw KubernetesResourceException("Could not get all namespaces for sever ${client.masterUrl}", e)
```
jeffmaury: s/sever/server

src/main/kotlin/org/jboss/tools/intellij/kubernetes/tree/KubernetesTreeStructure.kt
```diff
@@ -36,31 +40,30 @@ class KubernetesTreeStructure(private val project: Project) : AbstractTreeStruct
...
-        try {
             return when(element) {
                 rootElement ->
                     arrayOf(
                         Categories.NAMESPACES,
                         Categories.NODES,
                         Categories.WORKLOADS)
                 Categories.NAMESPACES ->
-                    getResourceModel().getAllNamespaces().sortedBy { it.metadata.name }.toTypedArray()
+                    runSafely("Loading resources failed", "Could not get all namespaces for server ${getResourceModel().getClient().masterUrl}") {
```
jeffmaury: Could notget namespaces to be coherent with the error for pod for example

src/main/kotlin/org/jboss/tools/intellij/kubernetes/utils/Notification.kt
```diff
@@ -0,0 +1,23 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
```
sbouchet: should be 2020
```suggestion
 * Copyright (c) 2020 Red Hat, Inc.
```

sbouchet:
@adietish : there is also notifications in Eclipse IDE, for odo explorer by example, but only for information purpose, like component creation/deletion and so on. Like @jeffmaury pointed out, errors should not be using notification, but proper logging.

adietish:
@jeffmaury please re-review

src/main/kotlin/org/jboss/tools/intellij/kubernetes/utils/Notification.kt
```diff
@@ -0,0 +1,23 @@
+/*******************************************************************************
```
jeffmaury: Is this class still used ?

adietish: No, it's not used any more. I thought about keeping it for future use. But then, non-used code is dead code, should be removed. I'll remove it.

Issue #6, opened 2020-01-10
support resources that are not namespaced


adietish:

adietish:
impl'd with PR at #23. Closing.

Issue #3, opened 2020-01-10
support other resources than just pods


adietish:

adietish:
depends on #5

adietish:
impl'd with PR at #23

PR #23, opened 2020-03-06
removed Namespaceprovider, impl'd OpenShiftCluster (#21)

### LINKED ISSUE #21: Support openshift clusters
Currently, we're only supporting plain kubernetes clusters (ex. minikube). We should also support OpenShift clusters.

adietish:
fixes #21, #6, #3


adietish:
@jeffmaury, @sbouchet : please review

steps:
1. ASSERT: have an openshift cluster that you can use
1. ASSERT: login into it via oc
1. EXEC: launch plugin

Result:
Resources tree should show the namespaces and pods of the cluster

1. ASSERT: keep plugin running
1. EXEC: create some pods (ex. create an app via template or via kubectl ```kubectl run ngnix --image=ngnix```)

Result:
Pods should show up as they are created

sbouchet:
I tested it locally and i've got a comment about the "Use Namespace" action. 
It is available on every element of the tree, even on the pods. 
using it on pods leads to errors. 
also, i tested the action on Projects, and it shows correctly the pods under that project. But maybe the name should be modified ? 
last comment is when i used a OpenShift cluster, there is an error for the Namespaces element's child : "Could Not load Namespaces". only pods and projects are displayed ( using remote 3.X and local 4.x OpenShift ) 

For minikube, it's working nice.

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/ResourceException.kt
```diff
@@ -0,0 +1,4 @@
+package org.jboss.tools.intellij.kubernetes.model
```
sbouchet: please add missing copyright header

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/ResourceModel.kt
```diff
@@ -0,0 +1,102 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
```
sbouchet: please change to 2020

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/IResourcesProvider.kt
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 Red Hat, Inc.
+ * Copyright (c) 2020 Red Hat, Inc.
```
sbouchet: please change to 
```suggestion
 * Copyright (c) 2019-2020 Red Hat, Inc.
```

adietish:
@sbouchet: thx for testing.

> I tested it locally and i've got a comment about the "Use Namespace" action.
> It is available on every element of the tree, even on the pods.

yes, this is not setup yet so that it would only be visible for nodes where it's applicable (namely namespaces or projects). In this sense this is a new issue to file.

> using it on pods leads to errors.

yes, that's expected (see comment above)

> also, i tested the action on Projects, and it shows correctly the pods under that project. But maybe the name should be modified ?

I'd agree. That's an enhancement that we can file.

> last comment is when i used a OpenShift cluster, there is an error for the Namespaces element's child : "Could Not load Namespaces". only pods and projects are displayed ( using remote 3.X and local 4.x OpenShift )

That's expected. If you're not admin on an OpenShift cluster, you're not allowed to list namespaces. You can only list projects. In a prior evolution of this code I had the projects displayed instead of the namespaces in the very same category: "Namespaces". VSCode extension doesn't behave like this: it always displays namespaces in "Namespaces" and adds an additional category "Projects" that lists the projects if the cluster is an OpenShift cluster. I believe that this is the correct way to behave since you can list and choose namespaces if you're admin.
I'd think that this shows another required improvement: the error node should tell you why listing wasn't possible, it currently doesnt (the log contains the stacktrace which unveils the reason though). We should improve the text of those error nodes. Agree?

> For minikube, it's working nice.

Great.
@sbouchet Given that your errors are expected, things seem to behave as they should, right?

adietish:
> please update copyrights to add 2020 ;)

@sbouchet, @jeffmaury I'd prefer not to have to add the new years on each and every file every time we modify them, keeping the creation date. WDYT?

jeffmaury:
This is what we are (trying) to do on JBoss Tools, but I'm fine with the policy of having the creation year only. But as this plugin is not yet out, it should have a single year of 2020 on every file !!!!

sbouchet:
@adietish : thanks for the explanations. 
> @sbouchet Given that your errors are expected, things seem to behave as they should, right?

Yes, things behave as they should for now :)

adietish:
> But as this plugin is not yet out, it should have a single year of 2020 on every file !!!!

In theory the right on the copy "is born" when you create the file. But I don't mind really, I'd just love not to have to update the copyright header whenever I change it in a new year. Correcting all to 2020.

adietish:
@sbouchet, @jeffmaury correct the year in the copyright headers, added the missing one in ```ResourceException```. Please re-review.

adietish:
> > I tested it locally and i've got a comment about the "Use Namespace" action.
> > It is available on every element of the tree, even on the pods.
> 
> yes, this is not implemented yet. The action is visible to all nodes even if it doesnt make sense or even errors. In this sense this is a new issue to file.
> 
> > using it on pods leads to errors.
> 
> yes, that's expected (see comment above)

filed #27 

> 
> > also, i tested the action on Projects, and it shows correctly the pods under that project. But maybe the name should be modified ?
> 
> I'd agree. That's an enhancement that we can file.

included in #27 

> 
> > last comment is when i used a OpenShift cluster, there is an error for the Namespaces element's child : "Could Not load Namespaces". only pods and projects are displayed ( using remote 3.X and local 4.x OpenShift )
> 
> I'd think that this shows another required improvement: the error node should tell you why listing wasn't possible, it currently doesnt (the log contains the stacktrace which unveils the reason though). We should improve the text of those error nodes. Agree?
> 
filed #28

Issue #32, opened 2020-05-06
run IJ compatibility tests for PRs too


adietish:
in #30 we added github actions to validate PRs. In there we also added a CI test that checks compatibility to different intellij versions. Given that intellij is changing it's API very frequently we should also verify every PR for its compatibility with different intellij versions.

Issue #25, opened 2020-03-13
Allow extensions to contribute nodes to the tree


adietish:
We need to allow extensions to contribute nodes to the resource tree. The example at hand is a future tekton extension which would contribute a cathegory "Tekton Pipelines" with different sub-categories and nodes. 
Here's how this looks in vscode:
![image](https://user-images.githubusercontent.com/25126/76634242-523bfa80-6546-11ea-8e1c-cfe3aa8c6e4f.png)
![image](https://user-images.githubusercontent.com/25126/76634373-74ce1380-6546-11ea-881b-29e34af921ef.png)

adietish:
* Documentations for tekton pipelines are here: https://openshift.github.io/pipelines-docs/docs/0.10.5/index.html

* the tekton extension for the fabric8 client is here: https://github.com/fabric8io/kubernetes-client/tree/master/extensions/tekton

adietish:
vscode extension to display tekton pipeline nodes for an openshift cluster: https://github.com/redhat-developer/vscode-openshift-tools/blob/9eebc38b1af59acf4a9d7d32c7b7cabb1b9f9480/src/extension.ts#L212

the kubernetes extension for vscode knows 2 types of categories:
* GroupingFolder (contains ResourceFolders or nodes)
* ResourceFolder (contains nodes)

The extension declares a contributor that contributes a resource folder "ImageStreams" to the grouping folder "Workloads" if the cluster is an openshift cluster:
```
clusterExplorer.nodeSources.resourceFolder("ImageStreams", "ImageStreams", "ImageStream", "ImageStream").if(isOpenShift).at("Workloads")
```

adietish:
example that shows how intellij-tekton is using the tekton extension for the fabric8 client to retrieve pipeline runs: https://github.com/redhat-developer/intellij-tekton/blob/master/src/main/java/com/redhat/devtools/intellij/tektoncd/tkn/PipelineRunDeserializer.java

jeffmaury:
It's the exact opposite: because of a bug in the Fabric8 Tekton extension, we had to deserialize the JSON output by our own and not using the extension. But this is fixed in 4.9.0 so we should be good to use it now

Issue #27, opened 2020-03-20
"Use Namespace" action: only show on Namespaces, label it "Use Project" on Projects


adietish:
Currently the action "Use Namespace" is shown on every node in the resource tree. It should only be shown on namespace resources and label it "Use Project" when used on Projects.

discussed in https://github.com/redhat-developer/intellij-kubernetes/pull/23#issuecomment-601939348

PR #34, opened 2020-05-12
display all contexts present in kubeconfig (#19)

### LINKED ISSUE #19: Display several clusters/contexts
We currently only display a single cluster node. The extension for VScode on the other hand displays several clusters:
image

adietish:
fixes #19, depends on #25, #32 
depends on https://github.com/fabric8io/kubernetes-client/issues/2215

### DIFF
--- a/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/util/KubeConfigContexts.kt
+++ b/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/util/KubeConfigContexts.kt
@@ -30,7 +30,7 @@ class KubeConfigContexts {
 			if (config == null) {
 				return null
 			}
-			val context = KubeConfigUtils.getCurrentContext(config)
+			val context = KubeConfigUtils.getCurrentContext(config) ?: return null
 			return contexts.find { it.context.user == context.user
 					&& it.context.user == context.user
 					&& it.context.cluster == context.cluster }
--- a/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/util/KubeConfigContexts.kt
+++ b/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/util/KubeConfigContexts.kt
@@ -12,7 +12,7 @@ package org.jboss.tools.intellij.kubernetes.model.util
 
 import com.intellij.openapi.diagnostic.logger
 import io.fabric8.kubernetes.api.model.Config
-import io.fabric8.kubernetes.api.model.NamedCluster
+import io.fabric8.kubernetes.api.model.NamedContext
 import io.fabric8.kubernetes.client.Config.KUBERNETES_AUTH_TRYKUBECONFIG_SYSTEM_PROPERTY
 import io.fabric8.kubernetes.client.Config.KUBERNETES_KUBECONFIG_FILE
 import io.fabric8.kubernetes.client.internal.KubeConfigUtils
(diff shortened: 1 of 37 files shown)


adietish:
The tree now displays contexts as it is done in vscode (only the active context - an openshift cluster - still differs):

image

I think that the labels are far to long, the current format is: `<current namespace>/<host>/<user>.`
I think that we should only display the host and decorate user and namespace like it's done ex. in the project tree:

image

@jeffmaury, @sbouchet: thoughts?

jeffmaury:
The issue with only the host is that you will get several entries in the tree with the same value

adietish:
> The issue with only the host is that you will get several entries in the tree with the same value

but they'd be distinguishable by the username & namespace written right to it in lighter letters?

jeffmaury:
Yes but as it's less visible I'm afraid users will be confused

adietish:
> Yes but as it's less visible I'm afraid users will be confused

I believe that these lengthy labels with elusively structured parts (current namespace, host, user) are more confusing than if the different components were visually comprehensible as such (bcs represented differently) .

adietish:
> I believe that the labels are that long that you only grasp them if you carefully read them - which I think is bad UX. I believe that you should immediately grasp what cluster they concern in the first place. Then in a 2nd more careful look you should be able to distinguish the ones that concern the same cluster but different user/default namespace. 

Apparently those labels are being discussed in vscode. I extracted the suggested change for these labels to it's own issue at #42 and put it on hold until we have a decision in vscode and try to match what they'll provide.

adietish:
This PR will be considered finished as soon as the tests are fixed (done) and the current context is labeled correctly (not done yet)

adietish:
labelling of the active context is now corrected:

image

This PR is ready for review. 
@jeffmaury can you please review this?

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/util/KubeConfigContexts.kt
```diff
@@ -0,0 +1,127 @@
+/*******************************************************************************
```
jeffmaury, line 1: My concerns:

- I think this is a general concern utility so I would see it in intellij-common
- If the env var contains several files, my understanding is that aggregation is performed, I am wrong ? @joshuawilson  may help there

adietish: > * I think this is a general concern utility so I would see it in intellij-common

This class is just a temporary solution. I want to contribute changes in kubernetes-client and think that this class wont be required any more. If there are parts which wont make it to the client library, I'll move it to intellij-common for sure.

joshuawilson: I just loaded 2 different kubeconfig files in 2 different locations. They did not merge. I can switch between them with `oc login [CLUSTER_URL]` but only one is active at a time.

adietish: @jeffmaury, @joshuawilson: my code that is switching to a kubeconfig that's provided in the env var is copied from kubernetes-client. There's no functional change intended compared to what the client library provides. 
I don't see any merging happening but an override instead. If there's an env var that provides a kubeconfig it'll be used, if there's none, the default kubeconfig in the home directory is what's serverd.

adietish: What we also discussed is that switching the project in an OpenShift cluster doesn't work. That's intended and is adressed in #27

PR #41, opened 2020-05-19
offer "Set as Current Cluster" in context menu for cluster that's listed (#40)

### LINKED ISSUE #40: offer "Set as Current Cluster" in context menu for cluster that's listed
The action should allow to switch the current cluster. The tree then removes the resources that are shown for the current cluster and displays them for the newly selected cluster.
Here's how it looks in vscode:
image

This depends on #19

adietish:
fixes #40


adietish:
in a 1st commit the tree had to be changed to display contexts, not clusters (switching clusters has to be done via activating a context). The tree now displays context as it is done in vscode (only the active context differs):
image
I think that the labels are far to long, the current format is: <current namespace>/<host>/user.
I think that we should only display the host and decorate user and namespace like it's done ex. in the project tree:
image

PR #46, opened 2020-05-29
display ImageStreams for OpenShift clusters (#45)

### LINKED ISSUE #45: subtask for #35: Display ImageStreams
image

is subtask of #35 (epic)

adietish:
fixes #45 

image

### DIFF
--- a/src/main/kotlin/org/jboss/tools/intellij/kubernetes/tree/OpenShiftStructure.kt
+++ b/src/main/kotlin/org/jboss/tools/intellij/kubernetes/tree/OpenShiftStructure.kt
@@ -12,16 +12,19 @@ package org.jboss.tools.intellij.kubernetes.tree
 
 import com.intellij.ide.util.treeView.NodeDescriptor
 import com.intellij.openapi.util.IconLoader
+import io.fabric8.openshift.api.model.ImageStream
 import io.fabric8.openshift.api.model.Project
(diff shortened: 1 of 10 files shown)


src/main/kotlin/org/jboss/tools/intellij/kubernetes/tree/OpenShiftStructure.kt
```diff
@@ -83,4 +94,11 @@ class OpenShiftStructure(model: IResourceModel): AbstractTreeStructureContributi
             model
     )
 
+    private class ImageStreamDescriptor(element: ImageStream, parent: NodeDescriptor<*>?, model: IResourceModel) : TreeStructure.Descriptor<ImageStream>(
+            element,
+            parent,
+            { element.metadata.name },
+            IconLoader.getIcon("/icons/project.png"),
```
jeffmaury, line 101: Not sure why this is the same as projects but noticed the VSCode one has no icons for ImageStreams

adietish: > Not sure why this is the same as projects but noticed the VSCode one has no icons for ImageStreams

I can remove the icons, there's no problem to that. I think that we should have good icons for all additional resources at some point. I just thought that I'd not care about icons for and take care of them altogether once we settled for good icons.

PR #82, opened 2020-06-30
Display 'Custom Resources' (#72)

### LINKED ISSUE #72: subtask to #35: display "Custom Resources"

adietish:
fixes #72


adietish:
@lstocchi can you please review this? There is 1 bug left that I am aware of: cluster scoped custom resources are not watched. This is most likely an issue in the client library that I'd like to resolve in a separate issue.

src/main/kotlin/org/jboss/tools/intellij/kubernetes/tree/KubernetesStructure.kt
```diff
@@ -604,10 +453,32 @@ class KubernetesStructure(model: IResourceModel) : AbstractTreeStructureContribu
...
+								.inAnyNamespace()
+								.list()
+								.sortedBy(resourceName)
+					}
+					parentElements { getRootElement() }
+				},
+				element<CustomResourceDefinition> {
+					anchor { it is CustomResourceDefinition }
+					childElements {
+						model.getCustomResources(it)
```
lstocchi, line 478: We should sort CustomResources in alphabetic order as in vscode

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/kubernetes/custom/GenericResourceFactory.kt
```diff
@@ -0,0 +1,87 @@
...
+	private const val GENERATION = "generation"
+	private const val NAME = "name"
+	private const val NAMESPACE = "namespace"
+	private const val RESOURCE_VERSION = "resourceVersion"
+	private const val SELF_LINK = "selfLink"
+	private const val UID = "uid"
+
+	fun createResources(items: List<Map<String, String>>): List<GenericResource> {
+		return items.stream()
+				.map { createResource(it) }
```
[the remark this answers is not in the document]
adietish: You're absolutely right, the type for items is erroneous, the correct one, which matches what one gets from the client (which is using jackson to deserialize the json into a map of <String, Object>) is 
```
fun createResources(items: List<Map<String, Any?>>): List<GenericResource>
```
My guess is that the generic parts of the map type are not used when casts are done at runtime and this thus succeeded at runtime. At compile-time this was fine for the kotlin compiler because of it is an up-cast String->Object.
Makes sense? Missing something?

src/main/kotlin/org/jboss/tools/intellij/kubernetes/tree/KubernetesDescriptors.kt
```diff
@@ -0,0 +1,245 @@
...
+			definition: CustomResourceDefinition,
+			parent: NodeDescriptor<*>?,
+			model: IResourceModel)
+		: Descriptor<CustomResourceDefinition>(
+			definition,
+			parent,
+			model
+	) {
+		override fun getLabel(element: CustomResourceDefinition): String {
+			return if (element.spec.names.singular.isNotBlank()) {
```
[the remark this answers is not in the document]
adietish: good catch, thx!

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/kubernetes/custom/GenericResourceFactory.kt
```diff
@@ -0,0 +1,94 @@
...
+			return ObjectMetaBuilder().build()
+		}
+		return ObjectMetaBuilder()
+				.withCreationTimestamp(metadata[CREATION_TIMESTAMP] as? String?)
+				.withGeneration(metadata[GENERATION] as? Long?)
+				.withName(metadata[NAME] as? String?)
+				.withNamespace(metadata[NAMESPACE] as? String?)
+				.withResourceVersion(metadata[RESOURCE_VERSION] as? String?)
+				.withSelfLink(metadata[SELF_LINK] as? String?)
+				.withUid(metadata[UID] as? String?)
```
lstocchi, line 65: I think it is enough to just write `metadata[UID] as? String` or `metadata[UID] as String?`. We should stick to using the same rule everywhere. For instance, above, you used `item[SPEC] as? Map<String, Any?>)`

adietish: IMHO are not the same thing. `metadata[UID] as? String` would return a `null` value if the cast fails (and not throw a class cast exception - see https://kotlinlang.org/docs/reference/typecasts.html#safe-nullable-cast-operator). `metadata[UID] as String?` would throw a class cast exception if the value is not null but of a different type.

jeffmaury (approved):
Worked for me. The only concern is that it does not work if you are not cluster admin but the VSCode extension has the same issue. And watched worked for cluster scoped resources: using kubectl I created a new ClusterTask and it was reflected in the tree

adietish:
> The only concern is that it does not work if you are not cluster admin but the VSCode extension has the same issue

Thanks for confirming. I filed this issue to #85

PR #84, opened 2020-08-11
dont use inexistent current namespace (#49)

### LINKED ISSUE #49: View seems to select wrong context
I have a strange configuration but worth testing.
In my kubeconfig I have 2 distincts contexts defined for the same cluster and user but pointing to different projects. One of them does not exists anymore.
So the context are:

- `demojeff/clusterurl/token`
- `jbosstools/clusterurl/token`

demojeff project does not exist anymore on the cluster.
Although my kubeconfig says the current-context is `jbosstools/clusterurl/token`, the view selects `demojeff/clusterurl/token`
And selecting `jbosstools/clusterurl/token` as the current context does not seem to work

adietish:
fixes #49 
depends on #72


adietish:
@jeffmaury this is now ready for review. Please review.

PR #87, opened 2020-08-21
watch kube config and update model & tree (#24)

### LINKED ISSUE #24: Track changes in ~/.kube/config and update resource tree
The plugin should update it's resource tree upon changes done to ~/.kube/config via kubectl, oc, etc.

adietish:
fixes #24

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,6 +48,17 @@ configurations {
     integrationTestRuntimeOnly.extendsFrom testRuntimeOnly
 }
 
+test {
+    afterSuite { desc, result ->
+        if (!desc.parent)
+            println("${result.resultType} " +
+                    "(${result.testCount} tests, " +
+                    "${result.successfulTestCount} successes, " +
+                    "${result.failedTestCount} failures, " +
+                    "${result.skippedTestCount} skipped)")
+    }
+}
+
 intellij {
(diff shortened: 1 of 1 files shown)


adietish:
@jeffmaury, @lstocchi: please review

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/Contexts.kt
```diff
@@ -73,10 +76,36 @@ class Contexts(
...
+		clear(false)
+		observable.fireModified(this) // invalidates root bcs there's no tree node for this
+	}
+
+	private fun closeCurrent() {
+		closeCurrent(true)
+	}
+
+	private fun closeCurrent(notify: Boolean) {
+		if (current == null) {
```
lstocchi, line 100: Is this check necessary? If this.current is null, its null value will be given to `val current` then the `return` will be made at `current?.close() ?: return`, no?

lstocchi (approved):
LGTM. Tested on CRC and it worked.

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/Contexts.kt
adietish: great catch, thx!

PR #95, opened 2020-10-05
replace cached resource if watch notifies that it changed (#92)

### LINKED ISSUE #92: Pods are not refreshed when they change status
steps:
1. EXEC: via kubectl or console: create a pod 
2. ASSERT: pod shows up in the list of pods Workloads > Pods with a red dot (it's not running yet)
3. EXEC: wait until pod gets "running"
4. ASSERT: pod should get a green dot

adietish:
depends on #91 which has to be merged first
fixes #92


adietish:
@jeffmaury, @lstocchi this is ready for review but #91 should be merged first. I thus keep this as a draft until the latter has been merged.

adietish:
@Here's how things behave

adietish:
Closed this PR by mistake, reopened it.

jeffmaury:
The following scenario did not work for me:
- start crc
- expand the **Projects** and **Pods** nodes
- install **OpenShift Pipelines** operator
- as soon as the **openshift-pipelines** project is displayed, set it as current
- the Pods node seems to be updated way after the pods are created (I compared by running `oc get pods -w -n openshift-pipelines` in a separate window and is not updated correctly: see what I got after several minutes:

image
Seems we are loosing events as 
- one pod has incorrect state
- pods for Triggers are missing

adietish:
@jeffmaury I can reproduce this. Investigating what's causing this.

adietish:
@jeffmaury, @lstocchi the above error is gone for me, pods that tekton is creating get updated and eventually turn green. 
Tests need to be fixed and additional ones created.
Please test.

adietish:
I found another bug. Steps:

1. ASSERT: freshly started plugin, have a 2 namespaces with 1+ pod
1. EXEC: expand Workloads > Pods
1. EXEC: switch namespace
1. ASSERT: pod in new namespace is listed
1. EXEC: kill the pod

Result:
List of pods is not refreshed, the pod is still listed even though it does not exist any more

adietish:
I found the issue and fixed it: when switching namespace, watches are stopped but were not restored. Next step: fix the tests & add tests so that they match the new logic.

adietish:
I tested against CRC and aws cluster: the change is fixing the issue for me. The tekton operator is creating 4 pods, which initially come up red and then turn green as they change to a running state.

adietish:
things now work for me: 
intellij-kubernetes-updating-tekton-pods
@jeffmaury @lstocchi please review

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/AbstractResourcesProvider.kt
```diff
@@ -10,45 +10,70 @@
...
+        logger<AbstractResourcesProvider<*>>().debug("Adding resource ${resource.metadata.name}.")
         // don't add resource if different instance of same resource is already contained
         synchronized(allResources) {
-            return allResources.find { areEqual(it, resource) } == null
-                    && allResources.add(resource as R)
+            val existing = allResources.find { resource.sameUid(it) }
+            return when (existing) {
+                resource -> false
+                null -> allResources.add(resource as R)
+                else -> replaceBy(existing, resource)
```
lstocchi, line 39: Why do you keep this else? When does `find` return a different value than a `resource` or `null`?

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/AbstractResourcesProvider.kt
```diff
@@ -10,45 +10,70 @@
...
         synchronized(allResources) {
-            return allResources.removeIf { areEqual(it, resource) }
+            val toReplace = allResources.find { resource.sameUid(it) } ?: return false
+            return replaceBy(toReplace, resource)
+        }
+    }
+
+    private fun replaceBy(toReplace: R, replaceBy: HasMetadata): Boolean {
+        val indexOf = allResources.indexOf(toReplace)
+        if (indexOf < 0) {
```
lstocchi, line 69: I guess you can just write `allResources[allResources.indexOf(toReplace)] = replaceBy as R`. This is a private method only called by the `replace` above and before you call this you already check if the element exists in the list otherwise return false. BTW not so importat

lstocchi (approved):
LGTM ... i think i'm not as good as Jeff at finding broken pods though :laughing:

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/AbstractResourcesProvider.kt
adietish: Wouldn't trying to set an element at a negative index throw an exception? In Java it would, not sure about kotlin yet.

adietish: Trying to set an element at a negative index throws:
```
java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 3
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
```
Missing something?

adietish: gotcha, you think that we could rely on the result of the find that we do right above. +1

lstocchi: What i meant was. By using `allResources.find { resource.sameUid(it) } ?: return false` you already verify if the element exists inside the collection (~index >=0) otherwise you never call the `replaceBy` function.

```
override fun replace(resource: HasMetadata): Boolean {
        if (!isOfCorrectKind(resource)) {
            return false
        }
        logger<AbstractResourcesProvider<*>>().debug("Replacing resource ${resource.metadata.name}.")
        synchronized(allResources) {
           ----> val toReplace = allResources.find { resource.sameUid(it) } ?: return false
            return replaceBy(toReplace, resource)
        }
    }
```

Inside the replaceBy you check again if index is greater than -1 but you already did this check above. It will never happen you will have a -1 as if the element doesn't exist this method will never be called, right? 
```
private fun replaceBy(toReplace: R, replaceBy: HasMetadata): Boolean {
        val indexOf = allResources.indexOf(toReplace) 
        if (indexOf < 0) { --> check if the element exists but you already made this check above
```

other thing is that when you use the indexOf you don't check the UID and i guess the check could return a false positive(???) bc maybe the object you save is different from the one you are checking.
I think it may happen 
```
Pod
uid: aaaa
normal config
``` 

then k8s does something

```
Pod
uid: aaaa
normal config
stuff added by k8s for some reason
```

I guess at this point the indexOf will fail but i'm just guessing.

lstocchi: I was writing while you already answered, sorry. Yes you got what i meant :laughing:. Give also a look at the indexOf, read above. Not sure how it works

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/AbstractResourcesProvider.kt
adietish: my understanding is the following:
* resource: it has found the very same instance, no need to replace it
* null: no object with the same uid was found
* else: a different instance with the same uid was found, we have to replace it. (ex. the watch has an update for a given object)

Maybe the case `resource` is very unlikely and could be dropped? Missing something?

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/AbstractResourcesProvider.kt
adietish: thinking twice I'd still prefer the current solution bcs it's more robust. The method would not rely on any external find that should happen so that it would not throw. At the same it, there's no perf hit. On the other hand I'd prefer not to have to find first and then look up again via index. I'd love to get the index when finding. Didnt' see how I'd achieve this.

lstocchi: Yeah, thinking about the indexOf it always works fine in the current use case as the object checked is the one found by 'find'. Resolving it then.

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/AbstractResourcesProvider.kt
lstocchi: No, i agree with your solution as well. It makes completely sense.

jeffmaury (approved):
CRC / OpenShift Pipelines is no ok on my conf

PR #104, opened 2020-10-15
Allow java consumers to consume (watch-) Supplier<T> (#102)

adietish:
fixes #102

PR #114, opened 2020-11-06
create custom resource provider when watching custom resource that wasn't loaded yet (#113)

### LINKED ISSUE #113: Custom resources kind is not watched if it already has instances
steps:

1. ASSERT: have custom resource definition existing on cluster
```
echo "apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: cars.openshift.pub
spec:
  group: openshift.pub
  names:
    kind: Car
    listKind: CarList
    plural: cars
    singular: car
  scope: Cluster
  subresources:
    status: {}
  version: v1" | kubectl apply -f - 
```
2. ASSERT: have custom resource instances that match the spec of this definition
```
echo "apiVersion: openshift.pub/v1
kind: Car
metadata:
  name: bmw
spec:
  date_of_manufacturing: "2014-07-01T00:00:00Z"
  engine: N57D30" | kubectl apply -f -
```
3. EXEC: launch plugin, unfold "Custom Resources"
4. ASSERT: your custom resource definition among the entries that are listed
5. EXEC: unfold your custom resource definition
6. ASSERT: it shows 1+ existing instances
7. EXEC: create a new instance of this definition
```
echo "apiVersion: openshift.pub/v1
kind: Car
metadata:
  name: alfa
spec:
  date_of_manufacturing: "2016-07-01T00:00:00Z"
  engine: CQ123456" | kubectl apply -f -

```
Result:
The tree wont get updated to show the new entry. Same is true when you delete your existing instance

Expected result:
The tree gets updated to show the new entry. The same is true for deleting the existing instance

adietish:
depends on #104 which has to be merged first
fixes #113

### DIFF
--- a/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/ResourceModel.kt
+++ b/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/ResourceModel.kt
@@ -31,6 +31,7 @@ interface IResourceModel {
     fun <R: HasMetadata> resources(kind: ResourceKind<R>): Namespaceable<R>
     fun resources(definition: CustomResourceDefinition): ListableCustomResources
     fun watch(kind: ResourceKind<out HasMetadata>)
+    fun watch(definition: CustomResourceDefinition)
     fun invalidate(element: Any?)
     fun addListener(listener: ModelChangeObservable.IResourceChangeListener)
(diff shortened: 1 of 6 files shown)


lstocchi (approved):
LGTM

Issue #109, opened 2020-11-06
NPE during Expanding pod under Nodes -> OpenshiftMaster element and calling a ContextMenu (right mouse click)


odockal:
Steps:
1. ASSERT: make sure you're on the "openshift" namespace
1. EXEC: expand Nodes > <node>
1. ASSERT: all existing pods are listed
1. EXEC: expand a node
1. ASSERT: Running state and IP children are expanded
1. EXEC: right click on running child

![image](https://user-images.githubusercontent.com/25126/99119787-18058a00-25fa-11eb-9853-bfa63783dd5d.png)

Result:
an error notification pops up reporting an error in the IDE. Looking into the log you'll find lots of these entries:
```
update failed for AnAction(org.jboss.tools.intellij.kubernetes.actions.UseNamespaceAction) with ID=org.jboss.tools.intellij.kubernetes.actions.UseNamespaceAction

java.lang.NullPointerException
	at com.redhat.devtools.intellij.common.actions.TreeAction.lambda$isVisible$3(TreeAction.java:83)
	at java.util.stream.MatchOps$1MatchSink.accept(MatchOps.java:90)
	at java.util.Spliterators$ArraySpliterator.tryAdvance(Spliterators.java:958)
	at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126)
	at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:498)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:485)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:230)
	at java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:196)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.noneMatch(ReferencePipeline.java:459)
	at com.redhat.devtools.intellij.common.actions.TreeAction.isVisible(TreeAction.java:83)
	at com.redhat.devtools.intellij.common.actions.TreeAction.update(TreeAction.java:68)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performDumbAwareUpdate(ActionUtil.java:171)
[...]

PR #122, opened 2020-11-16
added explanation to ResourcePropertyDescriptor#getElement (#109)

adietish:
'fixes' #109 (only adds javadoc explaining why ResourcePropertyDescriptor#getElement has to return null)


lstocchi (approved):
LGTM

PR #127, opened 2020-11-27
Replace Pod.isRunning() with PodStatusUtils.isRunning() in kubernetes-client (#126)

### LINKED ISSUE #126: Replace class Pod.isRunning() with PodStatusUtils.isRunning() available in kubernetes-client
When listing pods we're showing their status via a green or red dot in their icon. This was introduced in #54 which fixed #53. I then suggested this code to kubernetes-client in https://github.com/fabric8io/kubernetes-client/pull/2552 which is fixing https://github.com/fabric8io/kubernetes-client/pull/2293. Now that the contribution is available in kubernetes-client (v4.13.0) we should replace our code in intellij-kubernetes by the code in the client.

adietish:
depends on #153 which has to be merged first bcs it is using the kubernetes-client that contributes the replacements I contributes to the client.

fixes #126


adietish:
This should only be merged once intellij-tekton was also bumped to use kubernetes-client 4.13.0

lstocchi (approved):
LGTM

PR #129, opened 2020-11-30
compare cluster, namespace, user when checking if config changed (#123)

### LINKED ISSUE #123: Inacessible "Namespaces" are not refreshed when I start crc
Steps:
1. ASSERT: have a crc that is not started
1. EXEC: open "Namespaces" category 
1. ASSERT: "Namespaces" shows an error (expected)
image
1. EXEC: start crc and wait for it to be fully operational
1. EXEC: Refresh the tree

Result:
Nothing happens, "Namespaces" (or any other category) are not refreshed even though the cluster is now up and accessible

Expected result:
"Namespaces" should automatically refresh and display the items as soon as the server is up and running. At least a manual refresh should cause the namespaces to be displayed.

adietish:
depends on https://github.com/redhat-developer/intellij-common/pull/22 which has to be merged and published before.
fixes #123

### DIFF
--- a/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/KubeConfig.kt
+++ b/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/KubeConfig.kt
@@ -58,14 +58,12 @@ open class KubeConfig(private val refreshOperation: () -> Unit) {
 	}
 
 	protected open fun onConfigChange(watcher: ConfigWatcher, config: io.fabric8.kubernetes.api.model.Config) {
-		val oldCurrentContext = currentContext
-		val newCurrentContext = KubeConfigUtils.getCurrentContext(config)
-		val currentContextChanged = (oldCurrentContext != newCurrentContext)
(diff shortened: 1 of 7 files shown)


adietish:
@jeffmaury, @lstocchi: this should compile (checks in this PR successful) once intellij-commons was published with the change in https://github.com/redhat-developer/intellij-common/pull/22

src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/KubeConfig.kt
```diff
@@ -18,6 +18,7 @@ import io.fabric8.kubernetes.client.internal.KubeConfigUtils
 import io.fabric8.kubernetes.client.Config
 import io.fabric8.kubernetes.client.ConfigAware
 import java.nio.file.Paths
+import java.util.*
```
lstocchi, line 21: wildcard import

lstocchi (approved):
LGTM, tested with minikube and the namespaces node refresh correctly

PR #133, opened 2020-12-02
stop watching resources when tree node is collapsed (#132)

### LINKED ISSUE #132: Should stop watching resources when tree node is collapsed
We currently start watching resources when a tree node is expanded. We should now also stop watching resources when a node is collapsed.

adietish:
fixes #132

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceModel.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceModel.kt
@@ -33,14 +33,16 @@ interface IResourceModel {
     fun resources(definition: CustomResourceDefinition): ListableCustomResources
     fun watch(kind: ResourceKind<out HasMetadata>)
     fun watch(definition: CustomResourceDefinition)
+    fun stopWatch(kind: ResourceKind<out HasMetadata>)
+    fun stopWatch(definition: CustomResourceDefinition)
     fun invalidate(element: Any?)
     fun delete(resources: List<HasMetadata>)
     fun addListener(listener: ModelChangeObservable.IResourceChangeListener)
 }
 
 open class ResourceModel(
-        private val observable: IModelChangeObservable = ModelChangeObservable(),
-        private val contexts: IContexts = Contexts(observable, ::create)
+    private val observable: IModelChangeObservable = com.redhat.devtools.intellij.kubernetes.model.ModelChangeObservable(),
+    private val contexts: IContexts = Contexts(observable, ::create)
 ) : IResourceModel {
 
     override fun setCurrentContext(context: IContext) {
@@ -122,6 +124,14 @@ open class ResourceModel(
 
     override fun watch(definition: CustomResourceDefinition) {
         contexts.current?.watch(definition)
+    }
+
+    override fun stopWatch(kind: ResourceKind<out HasMetadata>) {
+        contexts.current?.stopWatch(kind)
+    }
+
+    override fun stopWatch(definition: CustomResourceDefinition) {
+        contexts.current?.stopWatch(definition)
     }
 
     override fun invalidate(element: Any?) {
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceWatch.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceWatch.kt
@@ -64,13 +64,13 @@ open class ResourceWatch(
         }
     }
 
-    fun ignoreAll(kinds: Collection<ResourceKind<*>>): Collection<ResourceKind<*>> {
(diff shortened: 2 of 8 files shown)


adietish:
in a first test this seems not to work: 
steps:
1. ASSERT: have a custom resource definition:
```
echo "apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: cars.openshift.pub
spec:
  group: openshift.pub
  names:
    kind: Car
    listKind: CarList
    plural: cars
    singular: car
  scope: Cluster
  subresources:
    status: {}
  versions:
  - name: v1alpha1
    served: true
    storage: false
  - name: v1beta1
    served: true
    storage: true
  names:
    kind: Car
    plural: cars" | kubectl apply -f -
```
2. EXEC: unfold "Custom Resources" > "cars"
3. EXEC: collapse "cars"
4. EXEC: repeat 2 & 3 several times and eventually have "cars" unfolded
5. EXEC: create a new car: 
```
$ echo "apiVersion: openshift.pub/v1beta1
kind: Car
metadata:
  name: chevrolet
spec:
  date_of_manufacturing: "2002-07-01T00:00:00Z"
  engine: CH654321" | kubectl apply -f -
```

Result:
There's no child popping up for "Custom Resources" > "cars"

Expected Result:
"Custom Resources" > "cars" should have a child popping up called "chevrolet"

adietish:
Works for me now. It can take a few sec before the new car appears though, because watching happens in the background, in a thread.

src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceModel.kt
```diff
@@ -124,6 +126,14 @@ open class ResourceModel(
         contexts.current?.watch(definition)
     }
 
+    override fun stopWatch(kind: ResourceKind<out HasMetadata>) {
+        contexts.current?.watch(kind)
```
lstocchi, line 130: Should this be stopWatch?

src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceModel.kt
```diff
@@ -124,6 +126,14 @@ open class ResourceModel(
...
+    }
+
+    override fun stopWatch(definition: CustomResourceDefinition) {
+        contexts.current?.watch(definition)
```
lstocchi, line 134: Should this be stopWatch?

lstocchi (changes requested):

src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceModel.kt
adietish: ouch, thanks for pointing out.

adietish:
@lstocchi verified, usecase works for me. Also further iterations where I collapse, unfold, create, delete etc. Please verify.

lstocchi (commented):
`Stop watching` works fine but i have a doubt (bc i had complains for this in IJ tekton as well). If you remove the watch and something changes on cluster, when you re-expand the node you don't see the actual situation. 
BTW the behavior it is the same as in vscode with this. 

E.g I expand the tree until secrets 
image

I create a new secret
```
apiVersion: v1
kind: Secret
metadata:
  name: test-1
type: Opaque
data:
  password: b3Bz
```
The tree is updated fine. Now i collapse the secrets node, then the configuration node.
I delete some secrets, create a new one. 
At this point if i re-expand the secrets node i don't see the real situation.

adietish:
> `Stop watching` works fine but i have a doubt (bc i had complains for this in IJ tekton as well). If you remove the watch and something changes on cluster, when you re-expand the node you don't see the actual situation.
gotcha, good catch!

I need to figure out why the following happens:

*Steps:*
1. ASSERT: have Configuration > Secrets expanded
2. EXEC: create secret:
```
echo "apiVersion: v1
kind: Secret
metadata:
  name: test-1
type: Opaque
data:
  password: b3Bz" | kubectl apply -f -
```
3. ASSERT: secret "test-1" appears
4. EXEC: collapse Secret, Configuration
5. EXEC: delete secret "test-1", create secret "test-2"
6. EXEC: expand Configuration > Secret

*Result:*
Secrets "test-1", "test-2" are listed

*Expected Result:**
Only secret "test-2" is listed.

adietish:
The reason for this is that the resource was deleted when the watch wasn't watchig (was stopped). When the "Secrets" node is expanded, the cache still contains the secret that was killed. It looks as if fixing this is simple: clear the cache when stopping the watch.

adietish:
@lstocchi fixed via clearing the cache when stopping the watch. Please re-review.

lstocchi (approved):
Awesome, works perfectly!! GJ!!

PR #137, opened 2020-12-11
allow to delete resources (#136)

### LINKED ISSUE #136: Should allow to delete resources
We should allow to delete resources like vscode allows users to:
image

adietish:
fixes #136


adietish:
@lstocchi test are missing, but the functionality is ready to be tested. Please review

adietish:
@lstocchi tests added, all done. Testing to see if I missed bugs.

lstocchi (commented):
There is really a lot to test so i'm going to split my testing to not get bored and writing comments along the way :laughing:
So far so good, only two things to mention.

The first is about the message the user get when he tries to delete resources. Too vague imo, we should atleast mention the name of the resources that are going to be deleted. In Tekton we have this kind of message, maybe it can give you an idea https://github.com/redhat-developer/intellij-tekton/pull/186

image


The second is a strange behavior. Not sure if you also have this. But with the pipelines resources (and only with them :scream: ) i see the image below when i delete a pipeline and the node never disappear. Do you have the same behavior?
image

jeffmaury:
That may be caused by tekton not emitting events

adietish:
> The first is about the message the user get when he tries to delete resources. Too vague imo, we should atleast mention the name of the resources that are going to be deleted. 

My bad, the code to display kind & name is in place but I forgot to include the name when refactoring it m-)
image

Force pushing together with a rebase.

adietish:
> The second is a strange behavior. Not sure if you also have this. But with the pipelines resources (and only with them 😱 ) i see the image below when i delete a pipeline and the node never disappear. Do you have the same behavior?
> image

@lstocchi I see the same with ex. "Persistent Volumes". As @jeffmaury pointed out I believe that events are not triggered, will have to investigate.

Weird enough I can't replicate this for pipelines:
1. EXEC: 
```
echo 'apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: pipeline-list-pipeline-repo-files
spec:
  resources:
    - name: pipeline-source-repo
      type: git
  params:
    - name: "path"
      default: "README.md"
  tasks:
    - name: list-pipeline-repo-files-1
      taskRef:
        name: list-pipeline-repo-files
      conditions:
        - conditionRef: "verify-no-file-exists-without-resource"
          params:
            - name: "path"
              value: "$(params.path)"
      resources:
        inputs:
          - name: optional-workspace
            resource: pipeline-source-repo' | kubectl apply -f -
pipeline.tekton.dev/pipeline-list-pipeline-repo-files configured
```
2. EXEC: delete the pipeline in "Custom Resources" > "pipelines" > "pipeline-list-pipeline-repo-files"

Result:
Pipeline (custom resource) disappears immediately.

What's different from your usecase?

adietish:
oc/kubectl use "Terminating" for a persistent volume where I use "(deleted)". See their code here:
https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/pkg/printers/internalversion/printers.go#L1685-L1687
```
func printPersistentVolume(obj *api.PersistentVolume, options printers.GenerateOptions) ([]metav1.TableRow, error) {
    ...
    if obj.ObjectMeta.DeletionTimestamp != nil {
        phase = "Terminating"
    }
    ...
```

adietish:
@lstocchi I assume that you were astonished to see "(deleted)" in the label to a resource and thus thought that it's not evident/intuitive enough. Oc/kubectl and the console both tag deleted resources as "terminated", I thus changed the labelling and now display **(terminaing)** (instead of "(deleted)"). Hope this helps. Thoughts?

ps. I can distinguish btw. resources that I deleted but didn't get a (watch-) notification yet, and the resources that I got notifications for. I currently dont, I label them both as "(terminated)".

lstocchi:
@adietish the thing that astonished me was to see the `deleted` tag. I was just expecting to not see the resource that i deleted after the tree refresh. But if there are cases when this may happen, not a problem for me. 

Regarding the term, for me `terminating` has a different meaning. You sent to the cluster a delete action and it is trying to terminate the resource but it still haven't succeeded. With persistent volume it may happen because you set up a pvc and the cluster is also trying to terminate that one without succeeding. 

In my case it was different, i could see the pipeline resource in deleted state in the IDE tree and it was actually deleted in the cluster. So if you change the label to `terminating` it changes the meaning. Not completely correct from my point of view but if they do this in the console, let's stay in sync with them. Even because there may be situation where the `terminating` label is better than the `deleted` one and if we don't have a way to recognize those states we cannot have a perfect solution.

So, ok for the `terminating` label. I'm going to keep testing this in the afternoon and let you know if i find something.

adietish:
> Regarding the term, for me `terminating` has a different meaning. You sent to the cluster a delete action and it is trying to terminate the resource but it still haven't succeeded. With persistent volume it may happen because you set up a pvc and the cluster is also trying to terminate that one without succeeding.

In your explanation there are 2 cases that I can label differently:
1. delete was sent but the cluster didn't send an event yet (modified, deleted, etc.)
2. delete was sent, cluster sent an event and is terminating 

I can use "(deleted)" in case 1. and "(terminated)" in case 2. Thoughts?

lstocchi:
> > Regarding the term, for me `terminating` has a different meaning. You sent to the cluster a delete action and it is trying to terminate the resource but it still haven't succeeded. With persistent volume it may happen because you set up a pvc and the cluster is also trying to terminate that one without succeeding.
> 
> In your explanation there are 2 cases that I can label differently:
> 
> 1. delete was sent but the cluster didn't send an event yet (modified, deleted, etc.)
> 2. delete was sent, cluster sent an event and is terminating
> 
> I can use **"(deleted)"** in case 1. and **"(terminating)"** in case 2. Thoughts?

Yes it sounds good to me :+1:

adietish:
pushed force bcs I rebased.
Impl'd what we discussed. Here's the code (#isWillBeDeleted returns true if the deletionTimestamp in the resource matches a marker that I set as soon as I execute deletion:
```kotlin
    open class ResourceDescriptor<T : HasMetadata>(
        ...
        override fun update(presentation: PresentationData) {
            super.update(presentation)
            if (isWillBeDeleted(element)) {
                presentation.presentableText += " (deleted)"
                presentation.setAttributesKey(CodeInsightColors.WRONG_REFERENCES_ATTRIBUTES)
            } else if (hasDeletionTimestamp(element)) {
                presentation.presentableText += " (terminating)"
                presentation.setAttributesKey(CodeInsightColors.WRONG_REFERENCES_ATTRIBUTES)
            }
        }
```
setting the marker when deleting the resource (`resources.forEach { setWillBeDeleted(it) }`):
```kotlin
abstract class ActiveContext<N : HasMetadata, C : KubernetesClient>(
    ...
    private fun delete(kind: ResourceKind<out HasMetadata>, scope: ResourcesIn, resources: List<HasMetadata>) {
        ...
        val deleted = provider.delete(resources)
        if (deleted) {
            resources.forEach { setWillBeDeleted(it) }
            modelChange.fireModified(resources)
         ...
    }
```

lstocchi (approved):
Tested and works fine for me.

PR #140, opened 2020-12-15
load non-namespaced cr in it's version with the highest priority (#139)

### LINKED ISSUE #139: Cluster scoped custom resources are displayed in the 1st available version (should be version with highest priority)
**Steps:**

1. EXEC: create cluster scoped custom resource definition with 2 versions (v1alpha1, v1beta1):
```
metadata:
  name: cars.openshift.pub
spec:
  group: openshift.pub
  names:
    kind: Car
    listKind: CarList
    plural: cars
    singular: car
  scope: Cluster
  subresources:
    status: {}
  versions:
  - name: v1alpha1
    served: true
    storage: false
  - name: v1beta1
    served: true
    storage: true
  names:
    kind: Car
    plural: cars
" | kubectl apply -f -
```
2. EXEC: create custom resource instance with v1alpha1
```
echo "apiVersion: openshift.pub/v1alpha1
kind: Car
metadata:
  name: alfa
spec:
  date_of_manufacturing: "2016-07-01T00:00:00Z"
  engine: CQ123456" | kubectl apply -f -
```
3. EXEC: change custom resource descriptor label so that it is displaying the version (it isn't without the change - you can of course also verify with a break-point):
Change [ResourceDescriptor](https://github.com/redhat-developer/intellij-kubernetes/blob/master/src/main/kotlin/org/jboss/tools/intellij/kubernetes/tree/TreeStructure.kt#L223) so that it shows the version (added method `postprocess`):
```kotlin
    open class ResourceDescriptor<T : HasMetadata>(
            element: T,
            parent: NodeDescriptor<*>?,
            model: IResourceModel) : Descriptor<T>(element, parent, model) {

        override fun getLabel(element: T): String {
            return element.metadata.name
        }

        override fun postprocess(presentation: PresentationData) {
            if (element == null) {
                return
            }
            presentation.addText(getLabel(element!!), SimpleTextAttributes.REGULAR_ATTRIBUTES)
            presentation.addText(" (${element!!.apiVersion})", SimpleTextAttributes.GRAYED_SMALL_ATTRIBUTES)
        }

    }
```
4. EXEC: launch the plugin, unfold **Custom Resources** > **cars** > **alfa**

**Result:**
`alfa`, the child for the custom resource definition `cars` is shown as **v1alpha1**. It should be shown as **v1beta1**

image

**Expected result:**
`alfa`, the child for the custom resource definition `cars` should be shown as **v1beta1**

image

adietish:
fixes #139

### DIFF
--- a/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/kubernetes/custom/NamespacedCustomResourcesProvider.kt
+++ b/src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/kubernetes/custom/NamespacedCustomResourcesProvider.kt
@@ -18,7 +18,7 @@ import io.fabric8.kubernetes.client.dsl.Watchable
 import io.fabric8.kubernetes.client.dsl.base.CustomResourceDefinitionContext
 import org.jboss.tools.intellij.kubernetes.model.resource.NamespacedResourcesProvider
 import org.jboss.tools.intellij.kubernetes.model.resource.ResourceKind
-import org.jboss.tools.intellij.kubernetes.model.util.getVersion
+import org.jboss.tools.intellij.kubernetes.model.util.createContext
 import java.util.function.Supplier
 
 class NamespacedCustomResourcesProvider(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/jboss/tools/intellij/kubernetes/model/resource/kubernetes/custom/NamespacedCustomResourcesProvider.kt
```diff
@@ -18,6 +18,7 @@ import io.fabric8.kubernetes.client.dsl.Watchable
 import io.fabric8.kubernetes.client.dsl.base.CustomResourceDefinitionContext
 import org.jboss.tools.intellij.kubernetes.model.resource.NamespacedResourcesProvider
 import org.jboss.tools.intellij.kubernetes.model.resource.ResourceKind
+import org.jboss.tools.intellij.kubernetes.model.util.createContext
 import org.jboss.tools.intellij.kubernetes.model.util.getVersion
```
lstocchi, line 22: idk why IJ doesn't remove it automatically but as the `GetVersion` is not used anymore this import can be removed.

lstocchi (approved):
Works fine. Tested with clusterTasks. GJ :+1:

PR #144, opened 2020-12-16
only show 'Set as Current Cluster' for non-active cluster (#141)

### LINKED ISSUE #141: Should not be able to "Set as Current Cluster" on current cluster
Steps:

1. ASSERT: have a kube config with at least 1 context
2. EXEC: select current context and pop up context menu & look into the items that are available

Result:
"Set as Current Cluster" is available and it shouldn't since we selected the current cluster 
image

Expected result:
"Set as Current Cluster" should not be avaiis available and it shouldn't since we're selecting the current cluster

adietish:
fixes #141


lstocchi (approved):
LGTM

Issue #147, opened 2020-12-22
should get informed which of the 2 namespaces could not be deleted


adietish:
Steps:
1. ASSERT: have a namespace that you created and that you can delete without harm (ex "test")
2. ASSERT: have the namespace "kube-system"
3. EXEC: select both and pick "Delete" in the context 

Result:
![image](https://user-images.githubusercontent.com/25126/102912864-3d4da980-447e-11eb-8046-fceaa2274242.png)
You're informed that both namespaces could not be deleted. In reality, only "kube-system" could not be deleted, "test" has been deleted.
There's actually no harm in this erroneous reporting. The namespace "test" disappears" as soon as it is removed by the cluster.

adietish:
follows up on #136 (PR #137)

adietish:
Of course the resources that succeeded should be additionally notified in an information notification as noted in #148

PR #151, opened 2020-12-23
Notify what resources have been deleted (#148)

### LINKED ISSUE #148: Should inform what resources had been deleted
follows up on #136 (PR #137)
After one deletes a resource, he should get informed what resources had been deleted. Vscode does that with a balloon:
image

adietish:
fixes #148

Introduces a notification that informs about the resources that have been successfully deleted:
image

The shortcoming here is of course like in #147: if deletion one or more resources could not be deleted, then all resources are reported as having failed, no notification of deleted resources happens. The expected result is that failed and succeeded delete operations should be notified separately. I'll fix that in #147


lstocchi (approved):
Awesome, this really improves the user experience. LGTM!!!

adietish:
> Awesome, this really improves the user experience. LGTM!!!

@lstocchi Yes, I agree. This is a nice enhancement.

PR #156, opened 2021-01-12
Create new context (& client) if token changes in kubeconfig  (#154)

### LINKED ISSUE #154: Non-reachable context should be refreshed (automatically) when I start crc
**Steps:**
1. ASSERT: have kubeconfig using an active context (cluster) that's not reachable/unresponsive (ex. crc is not running)
2. EXEC: launch plugin
3. ASSERT: active context is listed as such in the tool window: it has an arrow to it's left so that it can be unfolded
4. ASSERT: the active context has all the categories (Namespaces, Nodes, Workloads, etc.) but unfolding these leads to errors, of course
image
5. EXEC: make sure your context (cluster) is now reachable (ex. start crc)
6. EXEC: pick "Refresh" in the context menu of your active context

**Result:** 
Nothing happens, the context (cluster) is not refreshed.

**Expected Result:**
The context should be refreshed and list the available resources

PS. this use case cannot be emulated by logging out (`oc logout`) and back in again. Logout is changing the kubeconfig and thus this approach would cause 2 changes (1st change when doing logout, 2nd change when logging in back again) while the steps above only rely on a single change (login which is only replacing the token).

adietish:
depends on https://github.com/redhat-developer/intellij-common/pull/26, should be merged once that PR was merged and published
fixes #154

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ClientConfig.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ClientConfig.kt
@@ -19,7 +19,7 @@ import io.fabric8.kubernetes.client.DefaultKubernetesClient
 import java.nio.file.Paths
 import java.util.concurrent.Executors
 
-open class KubeConfig(private val refreshOperation: () -> Unit) {
+open class ClientConfig(private val refreshOperation: () -> Unit) {
 
 	private val executors = Executors.newFixedThreadPool(1)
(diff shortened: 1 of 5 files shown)


adietish:
@lstocchi, @jeffmaury: please review. Once this is merged, we're ready to release intellij-kubernetes 🕺

lstocchi (approved):
LGTM!! Let's go with the releaseee :man_dancing: :champagne:

jeffmaury (approved):

PR #161, opened 2021-01-27
renamed patchPluginXml>pluginDescription prop key (jetbrains constraint)

adietish:


jeffmaury (changes requested):
I tried to run ./gradlew buildPlugin and the resulting zip does not have a description in the plugin.xml

src/main/resources/META-INF/plugin.xml
```diff
@@ -2,6 +2,13 @@
     <id>com.redhat.devtools.intellij.kubernetes</id>
     <name>Kubernetes by Red Hat</name>
     <vendor email="developers@redhat.com" url="http://www.redhat.com">Red-Hat</vendor>
+    <description>
+        <![CDATA[
+        <h2>Overview</h2>
+        <p>A plugin for interacting with Kubernetes and OpenShift clusters.</p>
+        <p>The plugin provides functionalities and user experiences that are very close to the <a href="Kubernetes extension for vscode">https://marketplace.visualstudio.com/items?itemName=ms-kubernetes-tools.vscode-kubernetes-tools</a>.</p>
```
jeffmaury, line 9: I think this must be removed because it's very likely JetBrains will refuse it.

jeffmaury (changes requested):

src/main/resources/META-INF/plugin.xml
adietish: What is striking is that **"Overview"** is repeated as such when you search:
image
Then on the plugin page itself, "Overview" seems repeating what the page already states:
image
I'd thus remove "Overview" and drop the **icon** that also repeats the icon that's already displayed on top.

HTML per se seems supported as **links** that we have on other plugin pages are displayed:
image

Do you want me to remove the link to the vscode extension, is it what you mean?
 

Shouldn't we indicate a description in the plugin.xml?

adietish:
@jeffmaury dropped the icon, removed "Overview" and the link to the kubernetes extension for vscode. Please review.

src/main/resources/META-INF/plugin.xml
```diff
@@ -2,6 +2,12 @@
     <id>com.redhat.devtools.intellij.kubernetes</id>
     <name>Kubernetes by Red Hat</name>
     <vendor email="developers@redhat.com" url="http://www.redhat.com">Red-Hat</vendor>
+    <description>
+        <![CDATA[
+        <p>A plugin for interacting with Kubernetes and OpenShift clusters.</p>
+        <p>The plugin provides functionalities and user experiences that are very close to the Kubernetes extension for vscode (Visual Studio Code Kubernetes Tools).</p>
```
jeffmaury, line 8: You did not get me: I was about the reference to VSCode (concurrent IDE) and not the overview title !!!
Please restore it

jeffmaury (changes requested):

src/main/resources/META-INF/plugin.xml
jeffmaury: Add lt's see if the reference to VSCode pass the approval

adietish: I'd remove **Overview** and the icon as both repeat what the page already features (see screenshots & explanations above). I also think that referring to the vscode IDE should be included if possible. As such I'd submit as it currently is. Agree?

jeffmaury (approved):

PR #223, opened 2021-06-25
added toolbar to editor (#218)

### LINKED ISSUE #218: As a user I want to be able to force push
depends on #178 
based on comment in https://github.com/redhat-developer/intellij-kubernetes/pull/178#pullrequestreview-679456857

Especially in situations where I end up with changes in the editor but there's no push notification I'd would want to be able to manually trigger the push. Ex .via a editor toolbar with such an action.

adietish:
fixes #218


jeffmaury:
Added and modifying a label to a pod managed by knative gives me resource modified on the cluster after I pushed to the cluster

adietish:
@jeffmaury your issue is fixed for me. Hope it's for you now too. I introduced a mutex that prevents the change notification to process while the push is not finished. Please re-test

jeffmaury:
Seems to be better for me. However, I have 2 questions:
- when the document is changed, I have a notification to push to cluster, and there is a second dialog that asks for confirmation which I don't understand the value as I've already been asked a question
- I played with Kafka cluster and tried to add a new zookeeper node and then edit one of the zookeeper pods: while I was displaying the pod in editor, got a notification that the content was changed but the message is not quite clear as it refers to the previous revision:

image

Maybe we should update the message that content was updated from revision A to revision B

adietish:
> - when the document is changed, I have a notification to push to cluster, and there is a second dialog that asks for confirmation which I don't understand the value as I've already been asked a question

So in your understanding the confirmation dialogs when pushing or pulling should not appear as they basically duplicate the notification? Or is it that they should only appear if triggered by the actions (not when triggered with the links in the notifications)?

> - I played with Kafka cluster and tried to add a new zookeeper node and then edit one of the zookeeper pods: while I was displaying the pod in editor, got a notification that the content was changed but the message is not quite clear as it refers to the previous revision:
>
>image
>
> Maybe we should update the message that content was updated from revision A to revision B

Definitely sounds like a bug. Let me try to reproduce it. If the correct (updated) version was reported, is it still more clear to report version Y -> Y? I'd love to keep notification messages as short as possible if good if it doesn't harm info clarity.

adietish:
> @jeffmaury

> > -message is not quite clear as it refers to the previous revision:
> > Maybe we should update the message that content was updated from revision A to revision B
> 
> Definitely looks like a bug. Let me try to reproduce it. If the correct (updated) version was reported, is it still more clear to report version Y -> Y? I'd love to keep notification messages as short as possible if good if it doesn't harm info clarity.

Fixed it, it is now reporting the `resourceVersion` of the resource that was pulled (was reporting the version of the resource in the editor before pulling)

adietish:
@olkornii CI fails on ITs with the following error:
```
Run actions/upload-artifact@v1
  with:
    name: Linux-integration-test-reports
    path: build/reports/tests/integrationTest
  env:
    JAVA_HOME_8.0.292_x64: /opt/hostedtoolcache/jdk/8.0.292/x64
    JAVA_HOME: /opt/hostedtoolcache/jdk/8.0.292/x64
    JAVA_HOME_8_0_292_X64: /opt/hostedtoolcache/jdk/8.0.292/x64
Error: Path does not exist /home/runner/work/intellij-kubernetes/intellij-kubernetes/build/reports/tests/integrationTest
Error: Exit code 1 returned from process: file name '/home/runner/runners/2.278.0/bin/Runner.PluginHost', arguments 'action "GitHub.Runner.Plugins.Artifact.PublishArtifact, Runner.Plugins"'.
```
any idea what's wrong here?

jeffmaury:
> 
> 
> > @jeffmaury
> 
> > > -message is not quite clear as it refers to the previous revision:
> > > Maybe we should update the message that content was updated from revision A to revision B
> > 
> > 
> > Definitely looks like a bug. Let me try to reproduce it. If the correct (updated) version was reported, is it still more clear to report version Y -> Y? I'd love to keep notification messages as short as possible if good if it doesn't harm info clarity.
> 
> Fixed it, it is now reporting the `resourceVersion` of the resource that was pulled (was reporting the version of the resource in the editor before pulling)

OK but would be more clear if message would be: to revision XX

jeffmaury:
> 
> 
> @jeffmaury
> 
> > second dialog that asks for confirmation which I don't understand the value as I've already been asked a question
> 
> So in your understanding the confirmation dialogs when pushing or pulling should not appear as they basically duplicate the notification? Or is it that they should only appear if triggered by the actions (not when triggered with the links in the notifications)?

Yes in both cases seems like a duplication to me
> 
> > -message is not quite clear as it refers to the previous revision:
> > Maybe we should update the message that content was updated from revision A to revision B
> 
> Definitely looks like a bug. Let me try to reproduce it. If the correct (updated) version was reported, is it still more clear to report version Y -> Y? I'd love to keep notification messages as short as possible if good if it doesn't harm info clarity.

adietish:
@jeffmaury corrected both issues, please verify.

jeffmaury:
> 
> 
> @jeffmaury corrected both issues, please verify.

Works for me but I don't understand this:

image

adietish:
> > @jeffmaury corrected both issues, please verify.
> 
> Works for me but I don't understand this:
> 
> image

should definitely not happen, that's a bug. There should never be 2 notifications at once. Do you have steps to reproduce for me please?

jeffmaury:
> 
> 
> > > @jeffmaury corrected both issues, please verify.
> > 
> > 
> > Works for me but I don't understand this:
> 
> should definitely not happen, that's a bug. There should never be 2 notifications at once. Do you have steps to reproduce for me please?

Don't remember the exact step but added a new zookeeper pod to my Kafka cluster then edit the newly created Zookeeper node

adietish:
@jeffmaury provided fix for this, verifying that there's always only 1 notification aka "pulled" and "pull" notification dont appear simultaneously any more.

jeffmaury:
Retest and did not get notifications being duplicated.

jeffmaury (approved):
LGTM

PR #248, opened 2021-08-06
added telemetry (#212)

### LINKED ISSUE #212: Add telemetry information for the extension
Based on the guidelines for Red Hat tooling for IntelliJ, the extension should support telemetry reporting.

adietish:
fixes #212


build.gradle
```diff
@@ -115,10 +115,15 @@ test {
 intellij {
     version = ideaVersion //for a full list of Jetbrains IDEA releases please see https://www.jetbrains.com/intellij-repository/releases
     pluginName = 'Kubernetes by Red Hat'
-    plugins = ['yaml']
+    // plugins = ['yaml', '/Users/andredietisheim/Documents/jboss-workspaces/intellij-redhat-telemetry/build/idea-sandbox/plugins/Telemetry by Red Hat/' ]
```
jeffmaury, line 118: Should be removed

src/main/kotlin/com/redhat/devtools/intellij/kubernetes/settings/SettingsConfigurable.kt
```diff
@@ -0,0 +1,46 @@
+/*******************************************************************************
```
jeffmaury, line 1: Why do we have a settings ? The state is empty and it does not align with others plugins (ex OpenShift) ?

jeffmaury (changes requested):

src/main/kotlin/com/redhat/devtools/intellij/kubernetes/settings/SettingsConfigurable.kt
adietish: I dont get your point. Ex. tekton also has a `Configurable`. By the interface it looks as if it is the one responsible for creating the JPanel:
```
public @Nullable JComponent createComponent() {
```
What am I missing?

jeffmaury: Tekton has a configurable because it has specific options that are used by various actions (wizards,...) from the plugin. But the OpenShift plugin does not have a configurable and it telemetry enabled

adietish: @jeffmaury so you suggest to not have settings at all? I thought that you were complaining about the presence of the settings configurable. Sure can do that, there's no need for settings.

jeffmaury: Yes if there is no settings for the plugin then no configurable

build.gradle
```diff
@@ -112,10 +112,14 @@ test {
 intellij {
     version = ideaVersion //for a full list of Jetbrains IDEA releases please see https://www.jetbrains.com/intellij-repository/releases
     pluginName = 'Kubernetes by Red Hat'
-    plugins = ['yaml']
+    plugins = ['yaml', 'com.redhat.devtools.intellij.telemetry:0.0.2.18' ]
     updateSinceUntilBuild = false
 }
 
+runIde {
```
jeffmaury, line 119: The same should be added for integration tests see https://github.com/redhat-developer/intellij-quarkus/blob/main/build.gradle#L151

jeffmaury (changes requested):

build.gradle
adietish: done

jeffmaury (approved):
LGTM

PR #253, opened 2021-09-03
use IJ plugin version 1.1.4 (#252)

### LINKED ISSUE #252: Switch to IntelliJ Gradle plugin 1.x
The IntelliJ Gradle plugin has switched to 1.x. It has some incompatible changes but this is the maintained stream. See https://github.com/redhat-developer/intellij-openshift-connector

adietish:
fixes #252


build.gradle
```diff
@@ -113,19 +113,19 @@ test {
...
-    plugins 'yaml'
+    version = ideaVersion //for a full list of Jetbrains IDEA releases please see https://www.jetbrains.com/intellij-repository/releases
+    pluginName = 'Kubernetes by Red Hat'
+    plugins = ['yaml']
+    updateSinceUntilBuild = false
 }
 
 patchPluginXml {
-    sinceBuild sinceIdeaBuild
+    sinceBuild = "${sinceIdeaBuild}"
```
jeffmaury, line 123: Not sure to understand why a single = added does not work here, like other changes ?

jeffmaury (approved):
LGTM

build.gradle
adietish: here's the error without it:
> Gradle sync failed: Cannot set the value of task ':patchPluginXml' property 'sinceBuild' of type java.lang.String using an instance of type java.lang.Integer. (2 s 648 ms)

PR #262, opened 2021-09-09
use kubernetes-client 5.7.2

adietish:
we should bump to kubernetes-client >= 5.5.0 because it fixes an error that occurs when trying to fetch the version of an OpenShift cluster >= 4.6: https://github.com/fabric8io/kubernetes-client/issues/3189


src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/resource/kubernetes/CronJobsOperator.kt
```diff
@@ -14,7 +14,7 @@ import com.redhat.devtools.intellij.kubernetes.model.resource.NamespacedOperatio
 import com.redhat.devtools.intellij.kubernetes.model.resource.NamespacedResourceOperator
 import com.redhat.devtools.intellij.kubernetes.model.resource.ResourceKind
 import com.redhat.devtools.intellij.kubernetes.model.Clients
-import io.fabric8.kubernetes.api.model.batch.CronJob
+import io.fabric8.kubernetes.api.model.batch.v1beta1.CronJob
```
jeffmaury, line 17: Is this in sync with KubernetesDescriptors.kt ?

PR #264, opened 2021-09-15
Added intellij-common-ui-test-library and removed duplicated fixtures

olkornii:
Signed-off-by: Oleksii Korniienko <olkornii@redhat.com>


build.gradle
```diff
@@ -216,8 +208,14 @@ private def colorize(String color, String message) {
 }
 
 repositories {
-    mavenLocal()
+//    mavenLocal() // uncomment for testing with local maven
```
adietish, line 211: why would you disable using the local repo? This would force developers to uncomment when developing locally where one needs to deploy to the local repo and consume in plugin that depends on intellij-kubernetes.

build.gradle
```diff
@@ -216,8 +208,14 @@ private def colorize(String color, String message) {
...
     mavenCentral()
+    maven {
+        url 'https://repository.jboss.org/nexus/content/repositories/snapshots'
+    }
+    maven {
+        url 'https://repository.jboss.org/nexus/content/groups/public'
```
adietish, line 217: how's this repo different from the existing one at https://repository.jboss.org?

adietish (changes requested):

build.gradle
olkornii: I don't know why, but when I try to create a local intellij-common-ui-test-library and use it (mavenLocal () is uncommented), intellij-kubernetes won't build up. It works great with intellij-rsp, but not with kubernetes. This can be confusing, so I commented on this part, if someone uncomment it and project will not build, they will know why.

build.gradle
olkornii: You are right, we dont need it, it was in intellij-common-blabla manual, so I added it.

olkornii: I`ll remove it in the next commit.

build.gradle
adietish: would be important to find out why this happens. What's the error that you get?

adietish (approved):

PR #267, opened 2021-09-27
should only handle files on local filesystem (#258)

### LINKED ISSUE #258: InvalidPathException: Illegal char <:> at index 16: \Repository Diff: application.yml
Hi team!

My stacktrace: https://gist.github.com/RafaelPuerta/2f1130568e6a6f3047468c5f6fcbbfa1

Operation system : windows 10 version 21H1
IDE :Intellij Ultiimate 2021.2.1 Build #IU-212508055

adietish:
fixes #258

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ResourceFile.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ResourceFile.kt
@@ -10,15 +10,14 @@
  ******************************************************************************/
 package com.redhat.devtools.intellij.kubernetes.editor
 
-import com.intellij.diff.util.DiffUtil
 import com.intellij.openapi.application.ReadAction
 import com.intellij.openapi.application.WriteAction
 import com.intellij.openapi.util.io.FileUtil
 import com.intellij.openapi.util.io.FileUtilRt
+import com.intellij.openapi.vfs.LocalFileSystem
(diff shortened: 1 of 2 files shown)


adietish:
@jeffmaury very unsure if this fixes #258 since I cannot replicate it. My guess though is that the filesystem that the stacktrace for the issue is referencing isn't the local isn't the local one. So my check for it checks for the file protocol of the virtual file as one can see in IJ platform: 
[StandardFileSystems](https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/openapi/vfs/StandardFileSystems.java)
```java
private static final ClearableLazyValue<VirtualFileSystem> ourLocal = CachedSingletonsRegistry.markLazyValue(
    new ClearableLazyValue<VirtualFileSystem>() {
      @NotNull
      @Override
      protected VirtualFileSystem compute() {
        return VirtualFileManager.getInstance().getFileSystem(URLUtil.FILE_PROTOCOL);
      }
    }
  );
```

adietish:
@jeffmaury would love if you could test the editor on windows since I refactored it a bit, removing the instance variable with the path of the file, using the virtual file only (allows to remove quite a bit of code).

adietish:
@olkornii can you please have a look at why the ITs are failing?

jeffmaury (approved):
LGTM

PR #268, opened 2021-09-30
publish v0.2.3

adietish:
Signed-off-by: Andre Dietisheim <adietish@redhat.com>


jeffmaury (commented):
If I'm not mistaken, that would break existing plugins so I think we should have corresponding PRs in relevant repos

PR #271, opened 2021-10-04
allow to open local files (#217)

### LINKED ISSUE #217: Resource editor: should allow to edit files loaded from local disk (File > Open)
related to https://github.com/redhat-developer/intellij-kubernetes/pull/178#pullrequestreview-670158394

Should be able to push/reload files to/from the current cluster that one loads from the local filesystem.

adietish:
fixes #217

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ResourceFile.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ResourceFile.kt
@@ -95,7 +95,7 @@ open class ResourceFile protected constructor(
          * @param resource the resource that the file for should be returned for
          * @return the path for the given resource
          */
-        private fun getPathFor(resource: HasMetadata): Path {
+        fun getPathFor(resource: HasMetadata): Path {
             val name = getFilenameFor(resource)
             return Paths.get(TEMP_FOLDER.toString(), name)
         }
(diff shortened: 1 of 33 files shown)


adietish:
@olkornii this PR is introducing new behaviour when it comes to editor table titles:
* when editing a local file it shows the filename
* when editing a resource on the cluster, it shows name@namespace
image
There's no addendum "(1)" being added to the title any more. I updated the testing document accordingly.
Could I please ask you to fix the ITs so that they match the new behavior?

src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ResourceFile.kt
```diff
@@ -10,229 +10,172 @@
...
-        @JvmStatic
-        fun create(path: Path, virtualFile: VirtualFile? = null): ResourceFile? {
-            if (!isResourceFile(path)) {
-                return null
+            return ResourceFile(virtualFile)
+        }
+
+        private fun createTemporaryFile(resource: HasMetadata): VirtualFile? {
+            return try {
+                val file = FileUtilRt.createTempFile(resource.metadata.name, ".$EXTENSION")
```
lstocchi, line 70: Not sure if this is the behavior you want but let's say i open twice the same resource from the cluster, it will open two tabs with the same name. The only difference will be the tempfile because here it will be created a new one every call.
image

PR #312, opened 2021-12-01
ResourceEditorFactory#openEditor should use existing editor (#298)

### LINKED ISSUE #298: Modified (not pushed) editor is replaced without notice when cluster version is changed
**Steps:**
1. ASSERT: Have a statuefulset 
2. EXEC: "Edit..." statefulset 
3. ASSERT: editor with statefulset is opened
4. EXEC: change resource (ex. add label to a pod), **DON'T PUSH**
5. EXEC: in dashboard: edit same resource (ex. add a different label) & save it to cluster
6. EXEC: verify that editor is unchanged and that it is showing notification that allows to push (overwrite cluster) or pull (overwrite editor)

**Result:**
Editor is replaced with cluster version without any notification

**Expected result:**
Editor should stay unchanged and display a notification that allows to push (overwrite cluster) or pull (overwrite editor)

adietish:
fixes #298


jeffmaury (approved):
LGTM

PR #330, opened 2021-12-17
client now deserializes 'generation' property to Long, was Int (#324)

### LINKED ISSUE #324: Push notification displayed on resource opening
When i open a tekton task on cluster and this is opened in the editor, the push notification is shown. Nothing catastrophic but it is weird, i just opened a resource 8and there's no local change to it) there is no reason for me to push it.

It happens with all tekton resources but not with k8s native (e.g pods, daemonsets...)

image

adietish:
fixes #324


lstocchi (approved):
LGTM. It works fine now, thanks!!

PR #338, opened 2021-12-23
editor: should throw when retrieving crds fails (#337)

### LINKED ISSUE #337: Error looking up crds on stopped cluster is reported as "unsupported kind XXX/YYYY"
**Steps:**
1. ASSERT: have cluster stopped
2. EXEC: File > New > pipeline.yml
3. EXEC: paste the following into the editor that opens up:
```
---
kind: "Pipeline"
apiVersion: "tekton.dev/v1beta1"
metadata:
  creationTimestamp: "2021-12-17T15:25:35Z"
  generation: 3
  name: "app-deploy"
  namespace: "default"
  resourceVersion: "1794614"
  uid: "3a0f16ca-4c7b-4a89-8e0f-b10e620bc2ef"
spec:
  resources:
  - name: "pipeline-resource-type"
    type: "pipeline-type"
  tasks:
  - name: "smurf"
    taskRef:
      kind: "Task"
      name: "build-app"
```
**Result:**
Error notification shows up displaying "Unsupported resource kind Pipeline in version tekton.dev/v1beta1

**Expected result:**
Error notification should show "Error contacting cluster"

adietish:
fixes #337

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/EditorResourceFactory.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/EditorResourceFactory.kt
@@ -34,7 +34,7 @@ object EditorResourceFactory {
      * @return [HasMetadata] for the given editor and clients
      *
      */
-    fun create(editor: FileEditor, definitions: Collection<CustomResourceDefinition>): HasMetadata? {
+    fun create(editor: FileEditor, definitions: Collection<CustomResourceDefinition>?): HasMetadata? {
         return create(getDocument(editor), definitions)
     }
 
@@ -46,7 +46,7 @@ object EditorResourceFactory {
      * @return [HasMetadata] for the given editor and clients
      *
      */
-    fun create(document: Document?, definitions: Collection<CustomResourceDefinition>): HasMetadata? {
+    fun create(document: Document?, definitions: Collection<CustomResourceDefinition>?): HasMetadata? {
         return if (document?.text == null) {
             null
         } else {
@@ -67,7 +67,10 @@ object EditorResourceFactory {
      *
      * @see CustomResourceDefinitionMapping
      */
-    private fun create(jsonYaml: String, definitions: Collection<CustomResourceDefinition>): HasMetadata {
+    private fun create(jsonYaml: String, definitions: Collection<CustomResourceDefinition>?): HasMetadata? {
+        if (definitions == null) {
+            return null
+        }
         val resource = try {
             createResource<HasMetadataResource>(jsonYaml)
         } catch (e: RuntimeException) {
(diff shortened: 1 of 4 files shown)


lstocchi (changes requested):
Tried to replicate the error by following the example you had in the issue.
The problem for me is that the error message is not updated when the content changes.

E.g i opened a yaml file which contained a taskrun with `namespace: tekton-pipelines` and `generateName` field. Then i copy/paste the resource from your issue. The message i see it's related to the old resource i had. It doesn't get updated.

image

adietish:
> E.g i opened a yaml file which contained a taskrun with `namespace: tekton-pipelines` and `generateName` field. Then i copy/paste the resource from your issue. The message i see it's related to the old resource i had. It doesn't get updated.

Turns out that the error notification is actually updated with a new error, but the error that happens is always the same: the editor tries to load the crds and fails. Any further processing is then skipped.

jeffmaury (approved):
LGTM

lstocchi:
> @lstocchi:
> 
> > E.g i opened a yaml file which contained a taskrun with `namespace: tekton-pipelines` and `generateName` field. Then i copy/paste the resource from your issue. The message i see it's related to the old resource i had. It doesn't get updated.
> 
> Turns out that the error notification is actually updated with a new error, but the error that happens is always the same: the editor tries to load all the crds (so that it can find out if it knows the kind in the editor) and fails. Any further processing is then skipped. What tells you in what you see that this is the old error?

I would expect to read `.... with name: app-deploy in namespace: defaussslt` instead of `.... with name null in namespace tekton-pipelines`, no? :thinking:

adietish:
@lstocchi I dont understand why your error message is reporting fetching the crds in a specific namespace. The request is going to the cluster, not to a namespace. I suspected the following: I modeled the "crds" property as lazy kotlin property. Once called it would store the value and not request the value any more. I thus turned it into a hand made lazy value which would reset the value once an exception happens and then re-request it.
Please review again.

lstocchi:
> @lstocchi I dont understand why your error message is reports fetching the crds in a specific namespace. The request is cluster scoped, it is not namespaced. I suspected the following: I modeled the "crds" property as lazy kotlin property. Once called it would store the value and not request the value any more. I thus turned it into a hand made lazy value which would reset the value once an exception happens and then re-request it. Please review again.

Ok i misunderstood what you wanted to achieve. Now i got it. My concern is about the message though. Why do you want to show [null] in message? Is there any use case different from this one when this is useful to be shown?
In this example, i read `Operation: [list] for kind [crd] with name [null] in namespace [null] failed `. Showing `with name [null] in namespace [null]` adds confusion from my point of view. Isn't it enough to show `Operation: [list] for kind [crd]  failed`?
Can't you add a check so that when the name is null do not show `with name [null]` and when the namespace is null remove `in namespace [null]`?

adietish:
@lstocchi I get your point but I'm reluctant to change this. The details shown are what's in the exception that's thrown by the client library. The client library has tons of different exceptions, some are good, some not at all. Interfering usually ends up fixing a particular case leaving lots of others non-optimal. That's one of the reasons for me to only present those in the details.

PR #340, opened 2021-12-23
fire resource model changed if client config is changing (#339)

### LINKED ISSUE #339: Start/Stop of cluster does not cause tree to change
**Steps:**
1. EXEC: have plugin & minikube running, have some nodes expanded
2. EXEC: stop minikube

**Result:**
There's no change in the tree, it still shows the current cluster (which went unreachable) and the resources expanded.

**Expected:**
Tree should only show all contexts, none should be expanded (since none is reachable)

**Steps:**
1. ASSERT: start the plugin with a stopped minikube, all contexts are listed, none can be expanded (no expansion-arrows available)
2. EXEC: start minikube

**Result:**
There's no change in the tree, all the contexts are listed, none may be expanded. The plugin didn't notice that minikube set the current context and it went reachable.

**Expected Result:**
Once minikube is started the tree should show the current context and should be expandable and the resources browsed

adietish:
fixes #339

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/Contexts.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/Contexts.kt
@@ -10,6 +10,7 @@
  ******************************************************************************/
 package com.redhat.devtools.intellij.kubernetes.model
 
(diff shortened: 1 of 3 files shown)


jeffmaury (approved):
LGTM

PR #341, opened 2021-12-24
dont replace document after push, dont pull automatically when no local changes (#319)

### LINKED ISSUE #319: As a user I want to control if my local file is overwritten from the cluster pull
There is a lot of loaded configuration from the server that is not in my local file and I don't want to see it unless there was an option to diff with the cluster.

I certainly want a way to prevent my local file from being clobbered when I open it and the changes are pulled from the cluster.

adietish:
fixes #319


adietish:
@lstocchi, @jeffmaury: this PR can now be tested. I am now not touching the editor content at all. I keep track of the resource version (that i need to know whether the editor can be pulled from cluster) in a file attribute for the editor file.

lstocchi:
First impression. Played with it a bit and it works fine. The only thing i don't like much is to see the toolbar popping up after a push.
The scenarios:
If i open a new file in editor, add a new resource and push it, it works fine. The editor doesn't change and the toolbar is not shown.
If i open an existing resource from the tree, edit it on editor and push it, the toolbar is shown to ask if i want to pull the new version. But i've just pushed it.... puzzled face.
I still think that the toolbar shouldn't be shown if meaningless infos have being changed (e.g version after a push). In some way we should handle it. 
test-push

adietish:
> If i open an existing resource from the tree, edit it on editor and push it, the toolbar is shown to ask if i want to pull the new version. But i've just pushed it.... puzzled face. I still think that the toolbar shouldn't be shown if meaningless infos have being changed (e.g version after a push). In some way we should handle it. 
If I get your right you mean the notification by "tooldbar". The toolbar actually is the horizontal bar on top of the editor with the 2 icons. The thing that appears informing you of the possibility to push or pull is a notification (jetbrains terminus). 
I fully agree that the pull notification only should appear if your editor doesn't have the latest version. Is there a way for you to verify that the server did not create a new revision? If it is not, then this clearly is a bug. I'll try to reproduce it.

adietish:
@lstocchi can you please try to reproduce your issue? I tried but cannot. Maybe one of the several pushs I did on this day fixed it?

lstocchi (approved):
@adietish yeah sorry, idk why i keep calling it toolbar ... NOTIFICATION :laughing: 
BTW you were right, i was using old code apparently. Everything works perfectly, super super nice. I look forward to it be merged so i can use it on tekton :laughing: :clap: 
I found an error but it's not related to this. Opened a new issue for later https://github.com/redhat-developer/intellij-kubernetes/issues/345

LGTM

adietish:
@lstocchi, @olkornii fixed the bug found when testing: https://github.com/olkornii/intellij-kubernetes/pull/26/files#diff-4434da0a1603ed9571e2ab9baca10c4fac6a50536889049f327deed186903207R92
Please re-test.

lstocchi:
> For the editor these updates look like changes (the `resourceVersion` was bumped) and it thus displays the pull notification. The pull notification feels unexpected to the user even though is absolutely correct. I assume that we should therefore ignore changes that happen to the status field?

I noticed this behavior as well with `Jobs` but in that case the `spec` got modified by the cluster itself so the message was correct. 
I tried with Deployment but i don't see the notification so i guess you are already ignoring deployments changes. Looks nice. GJ!! Works fine for me!!

adietish:
@lstocchi: thx a lot for confirming and bringing in another meaningful usecase that demonstrates this.

> I noticed this behavior as well with `Jobs` but in that case the `spec` got modified by the cluster itself so the message was correct.

I plan to offer a diff/merge action so that the user can verify why he's notified, this should help to reduce the wth moment of those unexpected prompts. If the `spec` is altered by the cluster I tend to think that there's no way to refine the comparison editor/cluster to be more meaningful (I can ignore the clutter, the status but i dont think that I can ignore `spec` in a meaningful way.

PR #361, opened 2022-03-23
impl'd refresh of folder nodes (#359)

### LINKED ISSUE #359: "Refresh" (in context menu for tree item) of a category does nothing
related to #358 

I dont know how to test this because new projects, deployments, etc. would get added automatically when you create them and "Refresh" would be without visible effects. I know that refresh doesnt work when setting breakpoints.

adietish:
fixes #359


jeffmaury (approved):
LGTM but don't know how to test it

adietish:
@jeffmaury tried several approaches that could allow you to verify this but none of those worked. ex. in the sandbox logout in the web-ui does not invalidate your token but `oc logout` does. Unfortunately `oc` modifies the config file which the plugin then reacts to, altering the display before you can execute a refresh.

PR #378, opened 2022-04-08
ResourceEditor.isModified() should not compare resourceVersion  (#376)

### LINKED ISSUE #376: Push notification does not go away even though I just pushed
**Steps:**
1. ASSERT: make sure that you're connected to the Red Hat Sandbox (time consumed when pushing is higher than local cluster)
1. EXEC: open k8s resource yaml file in editor
ex. 
```
apiVersion: batch/v1
kind: Job
metadata:
  name: countdown
spec:
  template:
    metadata:
      name: countdown
    spec:
      containers:
      - name: counter
        image: centos:7
        command:
         - "bin/bash"
         - "-c"
         - "for i in 9 8 7 6 5 4 3 2 1 ; do echo $i ; done"
      restartPolicy: Never
```
2. EXEC: if resource doesn't exist yet, push it to create it on the cluster. 
3. EXEC: add label: `metadata > labels: jedi: yoda`
4. ASSERT: push notification appears prompting you to update the resource on the cluster ("update existing resource")
image
5. EXEC: hit "Push" link in the notification

**Result:**
Push notification is still displayed even though I just pushed. 
image

**Expected result:**
Push notification should be hidden and not displayed any more since I cluster & editor should be on the same page

adietish:
fixes #376

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ClusterResource.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ClusterResource.kt
@@ -18,7 +18,6 @@ import com.redhat.devtools.intellij.kubernetes.model.ResourceWatch
 import com.redhat.devtools.intellij.kubernetes.model.ResourceWatch.WatchListeners
 import com.redhat.devtools.intellij.kubernetes.model.util.isGreaterIntThan
 import com.redhat.devtools.intellij.kubernetes.model.util.isNotFound
-import com.redhat.devtools.intellij.kubernetes.model.util.isOutdated
 import com.redhat.devtools.intellij.kubernetes.model.util.isSameResource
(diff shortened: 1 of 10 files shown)


lstocchi (approved):
By following your example i still see the notification bar but that is due to the dynamic nature of a job resource i guess. This changes on cluster and differs from the local version. I don't think you can do much about it. By trying with another resource like secret it works fine.

LGTM

adietish:
> By following your example i still see the notification bar but that is due to the dynamic nature of a job resource i guess. This changes on cluster and differs from the local version. I don't think you can do much about it. By trying with another resource like secret it works fine.

yes, that's how I also look at what you describe. i can see the same with `Deployment`. The issue I tried to solve here is a wrong notification. The **PULL** notification is correct. In this issue I still had a **PUSH** notification.

PR #379, opened 2022-04-08
create resource before creating cluster watch (#345)

### LINKED ISSUE #345: apiVersion-error is not updated/removed when error is fixed
Below a gif that shows it. The steps to replicate it
- open a resource on editor. A pipeline in my case
- start changing the apiVersion to something which is wrong
- the notification shows up correctly telling you the apiVersion is wrong
- fix it by re-writing the correct apiVersion
- the notification bar doesn't disappear, it still shows the last wrong apiVersion

notification_bar

adietish:
fixes #345


jeffmaury (approved):
LGTM

Issue #304, opened 2021-11-19
Switch to main


jeffmaury:
Rename master branch to main as this is now the default

PR #390, opened 2022-04-25
updated github action to use main (#304)

adietish:


jeffmaury (approved):
LGTM. Please note that the definition of the Jenkins job needs to be updated as well, @sbouchet can help on this topic

sbouchet:
> LGTM. Please note that the definition of the Jenkins job needs to be updated as well, @sbouchet can help on this topic

default value of job needs to be updated. in the meantime, manual build can be trigerred using "main" as sha1 param value.

Issue #283, opened 2021-10-20
opened editor should switch to new current context if one switches the kube config


adietish:
depends on #217 which has to merged before one can experience this issue.

**Steps:**
1. ASSERT: have a yml file with a kubernetes resource opened
2. EXEC: change to a new context in your kube config (ex. switch from minikube to crc)

**Result:**
The editor has no change in notification. There's no "you can push to cluster".

**Expected result:**
Since I'm switching to a new context (namespace/cluster) where the resource in the editor does not exist yet a notification for pushing it to the new cluster should appear.

brunnels:
IMO I think the automatic notifications introduce a lot of complexity on the plugin implementation and a lot of noise/distraction for the end user.  I would want the ability to disable them so I can just use the always displayed push/pull/compare buttons.  

Specifically with this issue, when I click the push to cluster button, I think a warning should be displayed if I'm not in the current namespace notifying me of that and ask me if I wish to switch namespaces or cancel.

PR #404, opened 2022-05-10
use current context to manipulate editor resource (#283)

adietish:
Signed-off-by: Andre Dietisheim <adietish@redhat.com>


jeffmaury (commented):
LGTM but if you switch to a dead context (eg crc-admin or crc-developer while crc is down) you have a Connnection reset notification is that on purpose ? 

Otherwise LGTM

adietish:
> if you switch to a dead context (eg crc-admin or crc-developer while crc is down) you have a Connnection reset notification is that on purpose ?

It's not on purpose, I'll look into that. Definitely an issue for a fix-release 0.5.1.
Thanks for reviewing!

PR #411, opened 2022-06-01
wrap error details in HTML so that text wraps (#336)

### LINKED ISSUE #336: As a user i would like to see full error message
If i try to push a resource which is not valid, the error displayed is cut to a specific length.
In my case the error was 

```
namespaces/tekton-pipelines/tasks. Message: admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: 
validation failed: missing field(s): spec.steps[1].Image. Received status: Status(apiVersion=v1, code=400, details=null, kind=Status,
 message=admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: missing field(s): 
spec.steps[1].Image, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, 
additionalProperties={}), reason=BadRequest, status=Failure, additionalProperties={}).
```
and i can easily understand i forgot to add the image field.

But the plugin only shows 
image

It would be nice if the full error is pasted in the event log panel or if we add a link `show all` in the tooltip which gives the user the chance to read all the error message

adietish:
fixes #336

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ResourceEditor.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ResourceEditor.kt
@@ -38,7 +38,6 @@ import com.redhat.devtools.intellij.kubernetes.model.util.isGreaterIntThan
 import com.redhat.devtools.intellij.kubernetes.model.util.runWithoutServerSetProperties
 import com.redhat.devtools.intellij.kubernetes.model.util.toMessage
 import com.redhat.devtools.intellij.kubernetes.model.util.toTitle
-import com.redhat.devtools.intellij.kubernetes.model.util.trimWithEllipsis
 import io.fabric8.kubernetes.api.model.HasMetadata
 import io.fabric8.kubernetes.client.KubernetesClient
 import io.fabric8.kubernetes.client.utils.Serialization
@@ -160,7 +159,7 @@ open class ResourceEditor(
                     hideNotifications()
                     errorNotification.show(
                         toTitle(e),
-                        trimWithEllipsis(toMessage(e), 300) ?: ""
+                        toMessage(e)
                     )
                 }
             }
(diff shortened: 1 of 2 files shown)


adietish:
Here's how my 2-line error message looks:
grafik

@lstocchi: would be great if you could either test this or provide me the steps for your overly lengthy error message.

lstocchi:
> @lstocchi: would be great if you could either test this or provide me the steps for your overly lengthy error message. Would be great if we could verify how this would get displayed.

Just try pushing this 

```
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: add-task-2
spec:
  params:
    - name: first
      description: the first operand
    - name: second
      description: the second operand
  results:
    - name: sum
      description: the sum of the first and second operand
  steps:
    - name: add
    - name: sub
    - image: ok
    - image: no
```

The example above is a little extreme but it's what i'm afraid about.
The two lines print could solve our issue in simpler cases (E.g. if i only have 1 wrong step the core of the error is at the beginning of the description and it is correctly displayed) but if one has a longer error message we could end up with the same visualization problem. The important part, what i need to fix my error, may not been displayed.

Thinking aloud... wouldn't it be possible to print the full error in some tab in `problems tool window`/`event log`/`whatever` and have a link in the message tooltip that focuses on the tab with the error when clicked? Or adding a static message `to view full message opens up blabla...`?

adietish:
@lstocchi: tried your example and here's how it looks now that I removed any code that would trim the error message:
grafik
Would this work for you?

> Thinking aloud... wouldn't it be possible to print the full error in some tab in `problems tool window`/`event log`/`whatever` and have a link in the message tooltip that focuses on the tab with the error when clicked? Or adding a static message `to view full message opens up blabla...`?
I like the idea of displaying the error in the problems tab. I suggest though, that we file this to a separate issue.

lstocchi (approved):
Sorry i forgot to reply to your comment 5 days ago.

Tested it and LGTM. It is easy to understand the problem now :+1:

Issue #416, opened 2022-06-15
Fatal error when closing opened editor tabs


lstocchi:
**Steps:**
1. ASSERT: have 2 pods on the cluster
2. EXEC: open editors for both (double click in the resource tree)
3. EXEC: close the editors
4. EXEC: repeat 2-3 until a "fatal IDE error" notification appears

**Result:**
"fatal IDE error" notification is shown informing of an exception that was thrown when closing the editor. In the log you'll find the following:
```
java.lang.Throwable: Invalid file: file:///tmp/file-exists.yml (invalid, reason: File deleted: VfsEvent[deleted: file:///tmp/file-exists.yml]; VFPMI invalidated VFP during update: VfsEvent[deleted: file:///tmp/file-exists.yml]; stacktrace:
java.lang.Throwable
	at com.intellij.openapi.vfs.newvfs.impl.VirtualFileSystemEntry.appendInvalidationReason(VirtualFileSystemEntry.java:431)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualFileSystemEntry.invalidate(VirtualFileSystemEntry.java:422)
	at com.intellij.openapi.vfs.newvfs.persistent.PersistentFSImpl.invalidateSubtree(PersistentFSImpl.java:1682)
	at com.intellij.openapi.vfs.newvfs.persistent.PersistentFSImpl.executeDelete(PersistentFSImpl.java:1670)
	at com.intellij.openapi.vfs.newvfs.persistent.PersistentFSImpl.applyEvent(PersistentFSImpl.java:1487)
	at com.intellij.openapi.vfs.newvfs.persistent.PersistentFSImpl.processEvent(PersistentFSImpl.java:851)
	at com.intellij.openapi.vfs.newvfs.persistent.PersistentFSImpl.deleteFile(PersistentFSImpl.java:561)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualFileSystemEntry.delete(VirtualFileSystemEntry.java:261)
	at com.redhat.devtools.intellij.kubernetes.editor.ResourceFile$deleteTemporary$1.invoke(ResourceFile.kt:160)
	at com.redhat.devtools.intellij.kubernetes.editor.ResourceFile$deleteTemporary$1.invoke(ResourceFile.kt:159)
	at com.redhat.devtools.intellij.kubernetes.editor.ResourceFile.executeWriteAction$lambda-2$lambda-1(ResourceFile.kt:193)
[...]

PR #421, opened 2022-06-16
removed workaround required in kubernetes-client < 5.12.2 (#416)

### LINKED ISSUE #420: remove workaround that was required with kubernetes-client < 5.12.2
related to #370 

In #409 the kubernetes-client was bumped to 5.12.2. A workaround was forgotten that was required for the editor when using kubernetes-client < 5.12.2. It's now not required any longer. We should remove it.

adietish:
fixes #420


lstocchi (approved):
LGTM

PR #429, opened 2022-06-22
corrected handling of editor res w/o name (#427)

### LINKED ISSUE #427: Editing resource without name is reported as unreachable cluster
**Steps:**
1. ASSERT: have a yaml file with a resource that has no name:
```
apiVersion: v1
data:
  username: YWRtaW4=222
  password: MWYyZDFlMmU2N2Rm
kind: Secret
type: Opaque
```
2. EXEC: open it in an editor

**Result:**
Error notification is displayed telling you that the cluster could not be reached. In the details you can discover that the problem is that the resource has no metadata.
grafik

adietish:
fixes #427


adietish:
The correct error notification reports that the resource could not be fetched from the server. Astonishingly, a k8s resource without metadata is valid, but requesting it from server is only possible if it has a name. In the details then, the reason should be given: it has no name.

lstocchi (approved):
LGTM

PR #433, opened 2022-06-24
only consider left clicks for double clicks (#432)

### LINKED ISSUE #432: clicking left and then right on a resource in the tree opens it in an editor
**Steps:**
1. ASSERT: have at least 1 resource in the tree (ex. pod)
2. EXEC: click left on the resource and then right

**Result:**
The resource is opened in an editor and the context menu appears

**Expected result:**
The resource should not be opened in an editor, only the context menu should appear.

adietish:
fixes #432


lstocchi (approved):
LGTM. Works fine on Linux

PR #445, opened 2022-07-04
Issue 444

### LINKED ISSUE #444: Pull notification disappears when I bump the resourceVersion
**Steps:**
1. ASSERT: have no deployment `ngninx-rolling`
2. ASSERT: have no file `deployment-ngninx-rolling.yml` (that contains the deployment that follows)
3. EXEC: create a new file `deployment-ngnix-rolling.yml` with the following content
```
---
apiVersion: "apps/v1"
kind: "Deployment"
metadata:
  name: nginx-rolling
  resourceVersion: 1
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx-rolling
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 0
  template:
    metadata:
      labels:
        app: nginx-rolling
    spec:
      containers:
        - image: nginx
          name: nginx
          ports:
            - containerPort: 8080
```
4. EXEC: push it to the cluster
5. ASSERT: notification appears the deployment was changed on cluster
6. EXEC: modify **metadata > resourceVersion** to the value **42000000** (value needs to be >1Mio because of the server generated version that seems to be in this range)

**Result:**
notification that was displayed disappears

adietish:
fixes #444


lstocchi (changes requested):
In the case of pushing/pulling everything works fine because you work with the resource version sent back by the cluster but the possible problem arises when you use the resource version from the editor, edited manually by the user. Based on the documentation i found two interesting points.
1) Resource versions must be treated as opaque by clients and passed unmodified back to the server.
2) You must not assume resource versions are numeric or collatable. API clients may only compare two resource versions for equality (this means that you must not compare resource versions for greater-than or less-than relationships).

Regarding (1) i tried to edit the version locally and push it back to the cluster. Nothing was actually saved. The cluster ignores it. So in the `update` method (line 140 - https://github.com/redhat-developer/intellij-kubernetes/pull/445/files#diff-b583346c7a3dfd2787829cc6fbceda7f9ea0b332c70754c8372044d4abe175beR140 ) I think the saveResourceVersion (line 150) can be removed. The client only needs to trust the version returned by the server, so no need to update it. If the user edits it we can ignore it, no?

Regarding (2) in `ClusterResource.kt` to compare two versions the plugin uses `isGreaterIntThan`. Based on the doc we should just do `is not equal`. https://github.com/redhat-developer/intellij-kubernetes/blob/main/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ClusterResource.kt#L205

To replicate the error and see the wrong behavior of the two points above, just open your deployment and put a bigger resource version than the actual one. No notification bar is displayed.

adietish:
@lstocchi thanks a lot for looking up the specs. The questions regarding `resourceVersion` are coming up again and again since quite some time. The first step in the correct direction were to NOT send `uid` and `resourceVersion` when sending them to the server for storing. Here are the next steps obviously.

> Regarding (1) i tried to edit the version locally and push it back to the cluster. Nothing was actually saved. The cluster ignores it. So in the `update` method (line 140 - https://github.com/redhat-developer/intellij-kubernetes/pull/445/files#diff-b583346c7a3dfd2787829cc6fbceda7f9ea0b332c70754c8372044d4abe175beR140 ) I think the saveResourceVersion (line 150) can be removed. The client only needs to trust the version returned by the server, so no need to update it. If the user edits it we can ignore it, no?

Fully agree. I had this already impl'd in a PR but it went lost for some odd reason. 

> Regarding (2) in `ClusterResource.kt` to compare two versions the plugin uses `isGreaterIntThan`. Based on the doc we should just do `is not equal`. https://github.com/redhat-developer/intellij-kubernetes/blob/main/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ClusterResource.kt#L205
> 
> To replicate the error and see the wrong behavior of the two points above, just open your deployment and put a bigger resource version than the actual one. No notification bar is displayed.
> 

Fully agree here, too. Have to test throughoutly to make sure things still work fine after this change.

> P.S: the PR title must be updated 😆
done.

lstocchi (approved):
LGTM

>> Regarding (1) i tried to edit the version locally and push it back to the cluster. Nothing was actually saved. The cluster ignores it. So in the `update` method (line 140 - https://github.com/redhat-developer/intellij-kubernetes/pull/445/files#diff-b583346c7a3dfd2787829cc6fbceda7f9ea0b332c70754c8372044d4abe175beR140 ) I think the saveResourceVersion (line 150) can be removed. The client only needs to trust the version returned by the server, so no need to update it. If the user edits it we can ignore it, no?

> Fully agree. I had this already impl'd in a PR but it went lost for some odd reason.

Only missing point is this but i think you will have it in a diff PR

Issue #214, opened 2021-05-27
As a user I want to be able to open a terminal in a pod


adietish:
related to #215 

A user reported why he liked jetbrains kubernetes plugin with the following:
> It has even a nice "run shell inside the pod" feature (which of course doesn't work for containers without a shell, like the distroless builds done by ko)

PR #456, opened 2022-07-18
impl'd remove clutter action (#274)

### LINKED ISSUE #274: Editor: remove clutter from yaml/json
When pulling a resource into the editor a lot of 'clutter' is added to the yaml/json in the editor. We should allow the user to remove it. Ideally with an option for automatic removal in the preferences  

related to https://github.com/redhat-developer/intellij-tekton/issues/309

https://github.com/itaysk/kubectl-neat removes clutter from yaml and makes it more readable. We should provide similar functionality.

adietish:
fixes #274 
depends on https://github.com/redhat-developer/intellij-common/pull/153

Implements a basic action that removes clutter from k8s resource metadata. There's no restore nor disabling the action if there's no clutter (editor, initial resource etc.). You can still click the action but there's simply no effect.
I also didn't add it to the diff editor even though this could be very handy. I think that tracking the remote is neat but not that much of a need for the default dev.
I think that adding "shell to pod" is the better investement (a feature especially mentioned by users of the jetbrains plugin).
@jeffmaury, @lstocchi: Agree?


adietish:
@lstocchi, @jeffmaury: which one of the 2 icons would you vote for:
* https://github.com/redhat-developer/intellij-kubernetes/pull/456/files#diff-952b539cc1b8ba8bba72a6bdf6d4af3e66d85d97e3ad36b447c82649ee61c434
* https://github.com/redhat-developer/intellij-kubernetes/pull/456/files#diff-b33b5efe3b765f02306d22a78e4fb21fc7fa8eec621a7fa44b6607706730b57f

I vote for no2. You?

lstocchi:
> which one of the 2 icons would you vote for: I vote for icon1, you?
> 
> * 
> * 

ok icon1 for me as well

olkornii:
@adietish nothing, works for me localy and on GA: https://github.com/redhat-developer/intellij-kubernetes/runs/7496076081?check_suite_focus=true
I'll try to make them more stable, sorry for that :(

olkornii (approved):
@adietish nothing, works for me localy and on GA: https://github.com/redhat-developer/intellij-kubernetes/runs/7496076081?check_suite_focus=true (I have selected build from "Checks" and "Re-run failed jobs")
I'll try to make them more stable, sorry for that :(

adietish:
> @adietish nothing, works for me localy and on GA: https://github.com/redhat-developer/intellij-kubernetes/runs/7496076081?check_suite_focus=true (I have selected build from "Checks" and "Re-run failed jobs") 

great news :)

lstocchi (approved):
It works. So LGTM for the new action.
The only things that do not convince me are 
1) the button in the toolbar is not really visible. Looking at them (not just the button added here) they seem to be disabled.
This is the k8s toolbar 
image
This is an IJ toolbar
image

2) Maybe we could think adding the reverse action. I can remove clutters but what if i want to re-add them? The only way i have now is to close and re-open it.

3) Adding a setting to keep the `remove clutter` the default behavior could be something useful? So, if enabled, every time i open a resource, clutters are removed automatically.

adietish:
> It works. So LGTM for the new action. The only things that do not convince me are
> 
>     1. the button in the toolbar is not really visible. Looking at them (not just the button added here) they seem to be disabled.
>        This is the k8s toolbar
>        image
>        This is an IJ toolbar
>        image
> 

+1, will lighten them up and make sure I have separate icons for the light theme: #469 

>     2. Maybe we could think adding the reverse action. I can remove clutters but what if i want to re-add them? The only way i have now is to close and re-open the resource.

I agree that there's some benefit to be able to get the metadata back. Not sure how much it is though (how likely is it that a user wants these back?). Do you have some insights from telemetry maybe?
Undo works for me, I remve the metadata and hit Ctrl+Z right afterwards. I get them back. 
Then there's local file history, where I can get them back, it's not convenient though.
Lastly the diff would also allow you to get them back.
Not good enough?
  
> 
>     3. Adding a setting to keep the `remove clutter` the default behavior could be something useful? So, if enabled, every time i open a resource, clutters are removed automatically.
+1 on this, too. I created: #470

I also want to separate the action from the other 3 by a separator since it's not related to those. Those are all about interacting with the cluster: push, pull, diff. Remove clutter is only about the editor it should therefore be more clearly separated. Either by a separator or by putting it to the right corner of the toolbar. 
 
All in all I believe that there's more benefit in the feature "shell to pod" #214 and thus prioritize these enhancement with a lower prio.

Issue #472, opened 2022-08-15
Remove remote-robot


olkornii:
Remote-robot is included in intellij-common-ui-test-library, no need to have it in kubernetes.

olkornii:
sub-task for #296

PR #473, opened 2022-08-15
removed remote robot

olkornii:
Signed-off-by: Oleksii Korniienko <olkornii@redhat.com>

### DIFF
--- a/src/it/java/org/jboss/tools/intellij/kubernetes/BasicTests.java
+++ b/src/it/java/org/jboss/tools/intellij/kubernetes/BasicTests.java
@@ -41,6 +41,8 @@ import org.jboss.tools.intellij.kubernetes.tests.CreateAnotherTypeResourceByEdit
 
 import java.time.Duration;
 import java.util.List;
+import java.util.logging.Logger;
+import java.util.logging.Level;
 
 /**
  * JUnit UI tests for intellij-kubernetes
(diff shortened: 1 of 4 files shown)


olkornii:
relates to #472

src/it/java/org/jboss/tools/intellij/kubernetes/BasicTests.java
```diff
@@ -142,6 +142,7 @@ public static void closeGotItPopup() {
 
     public static void closeOpenedEditors() {
         List<SingleHeighLabelFixture> singleHeighLabelsList = robot.findAll(SingleHeighLabelFixture.class, byXpath("//div[@class='SingleHeightLabel']"));
+        System.out.println("Next opened editors will be closed: " + singleHeighLabelsList);
```
adietish, line 145: wouldnt it be better to log instead of sys-out? a log could be set to be printed to std out but also to a file?

olkornii: fixed

adietish (changes requested):
@olkornii looks like your commits somehow contains code from my PR for removing clutter (see ex. your PR would add changes to `RemoveClutterAction` and the icon).

adietish (approved):

PR #479, opened 2022-09-02
display pods for job (#478)

### LINKED ISSUE #478: resource tree: pods for jobs are not displayed
**Steps:**
1. EXEC: create a job:
```
apiVersion: batch/v1
kind: Job
metadata:
  name: endlessdate
spec:
  template:
    metadata:
      name: countdown
    spec:
      containers:
        - name: counter
          image: centos:7
          command:
            - "bin/bash"
            - "-c"
            - "for (( ;; )); do date +\"%H%m%s\"; sleep 2; done"
      restartPolicy: Never
```
2. EXEC: navigate to **Workloads > Jobs > enlessdate** and expand it

**Result:**
No pod for the job **endlessdate** is listed, it is displayed in **Pods** though

adietish:
fixes #478 

This PR fixes the bug for jobs and cronjobs. Cronjobs create jobs which in turn then show up in **Jobs** and their pods are displayed as children of the job that the `CronJob` created.
grafik

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/resource/Filters.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/resource/Filters.kt
@@ -10,16 +10,17 @@
  ******************************************************************************/
 package com.redhat.devtools.intellij.kubernetes.model.resource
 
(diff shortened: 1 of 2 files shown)


lstocchi (approved):
LGTM

PR #494, opened 2022-09-27
create new client when changing current namespace (#492)

### LINKED ISSUE #492: when switching current namespace, editor still saves to previous namespace
**Steps:**
1. ASSERT: have 2+ namespaces (ex. on minikube: "default" & "kube-public")
2. ASSERT: have "default" set as current namespace
3. EXEC: open the following in an editor:
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
    - name: nginx
      image: nginx:1.14.2
      ports:
        - containerPort: 80
```
4. EXEC: switch to "kube-public" namespace
5. EXEC: push editor to cluster

**Result:**
Pod is created in "default" even though I switched to the "kube-public" namespace

adietish:
fixes #492


adietish:
@jeffmaury, @lstocchi, @olkornii: I woulld love if you could test this change a bit while I fix the unit tests. This is a big change in all areas (model & tree) and I thus would appreciate a lot if I you could test that things still work as expected.

adietish:
@lstocchi, @jeffmaury: all done here, existing unit-tests were fixed new ones added. This is a big PR which affects many aspects of this plugin (tree, editor, etc.) so this should be tested throughly. Please review.

lstocchi (changes requested):
It seems that the tree doesn't refresh when some change happens on the cluster. 
I push a pod, i don't see it in the tree. By switching namespace, it eventually shows up. Now i delete it. Same thing. The tree is not refreshed even by clicking the refresh action

In the image you can see the message `nginx-3 was deleted` but it is visible in the tree
image

adietish:
@lstocchi:
> It seems that the tree doesn't get refreshed when some change happens on the cluster. I push a pod, i don't see it in the tree. By switching namespace, it eventually shows up. Now i delete it. Same thing. The tree is not refreshed even by clicking the refresh action

I cannot replicate your issue. Things work fine or me. Maybe your branch wasn't fully built before launching (happens regularly to me, I always `clean` before launching?
If this doesn't help, I'd kindly ask you to provide your pod yaml please.
I then would like to compare the minikube setup. What's your version? I'm using
```
minikube version: v1.27.1
commit: fe869b5d4da11ba318eb84a3ac00f336411de7ba
```

lstocchi:
I'm still able to replicate it even after clean.
It only happens after i start switching namespace.
The pod used is 
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-7
spec:
  containers:
    - name: nginx
      image: nginx:1.14.2
      ports:
        - containerPort: 80
```
my minikube
```
minikube version: v1.23.0
commit: 5931455374810b1bbeb222a9713ae2c756daee10

```

adietish:
@lstocchi: thanks for testing again. I found an issue where the plugin would not reflect a change in the config correctly. This for is/was related since any change of current namespace/context is save to the config file. I fixed and pushed the change. Can you please try it again?

adietish:
@lstocchi found and fixed the issue. Hope there are no others left that I missed. Thanks for reviewing again.

lstocchi (approved):
LGTM everything works fine for me

PR #505, opened 2022-10-21
balloon goes away on click outside & only use strictly req height (#491)

### LINKED ISSUE #491: Error balloon does not go away if I switch editor
follows up on #485

**Steps:**
1. ASSERT: have an editor opened
1. EXEC: open the following in an editor
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
    - name: nginx
      image: nginx:1.14.2
      ports:
        - containerPort: 80
```
2. EXEC: push it to the cluster 2x
3. ASSERT: 2nd push (or fist if the pod already existed) errors
4. EXEC: click on "Details" link
5. ASSERT: error balloon shows up
6. EXEC: (do not close the error balloon) switch to another editor

**Result:**
Error balloon is still visible even though it belongs to the editor that is now overlapped by the current editor.
grafik

adietish:
fixes #491


adietish:
2 things change with this PR:
* the balloon will close automatically if one clicks outside (that's the case when selecting a different editor)
* the balloon will only use the height that's required by the message in the balloon. It won't exceed a maximum of 400px though:
    * just the required height:     
image
    * not more than 400px:
    image

jeffmaury (approved):
LGTM

PR #510, opened 2022-10-27
display error in console tab & allow reconnect (#509)

### LINKED ISSUE #509: I dont want to close & reopen the terminal if connecting to the container fails
depends on #214
If you try to open a terminal to a pod whose container is not ready, I have to close the terminal and open a new one. I would like to be able to re-trigger opening the terminal.

adietish:
fixes #509

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/console/ConsoleTab.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/console/ConsoleTab.kt
@@ -11,27 +11,32 @@
 package com.redhat.devtools.intellij.kubernetes.console
 
 import com.intellij.execution.ui.ConsoleView
+import com.intellij.icons.AllIcons
 import com.intellij.openapi.Disposable
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.project.Project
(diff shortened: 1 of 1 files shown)


lstocchi:
Do any of you have a pod which is slow at starting and can be used for testing?

jeffmaury:
> Do any of you have a pod which is slow at starting and can be used for testing?

To test such a case, first ssh into minikube and delete the image (docker rmi xxx) and then create a new pod: the pod will be in that pending state the time the image is downloaded

lstocchi (commented):
I'll give it another try after it is rebased but LGTM

jeffmaury (approved):
LGTM but found this corner case.
I pod has 2 containers. So while the images are being loaded I open a terminal and UI tells me can connect to the container for the first open. When the pod is running, I click Reconnect and got a terminal for the first container. But when I switch to the second one, I have the same message where I expected to be connected to the container.

adietish:
> LGTM but found this corner case. I pod has 2 containers. So while the images are being loaded I open a terminal and UI tells me can connect to the container for the first open. When the pod is running, I click Reconnect and got a terminal for the first container. But when I switch to the second one, I have the same message where I expected to be connected to the container.

We tried to reproduce this when in a call and couldn't. I'll thus merge this PR and keep this possible error in mind, filing it when it happens and I've found out how to reproduce it.

Issue #502, opened 2022-10-17
"fatal IDE Error" (WebSocketHandshakeException) when trying to open a terminal to a container that is NOT running


adietish:
depends on #214 

**Steps:**
1. ASSERT: have a pod with a stopped coninter:
`alpine-container` terminates as soon as the bash script finished executing:
```
apiVersion: v1
kind: Pod
metadata:
  name: multi-pod2
spec:
  restartPolicy: Never
  volumes:
    - name: shared-data
      emptyDir: {}
  containers:
    - name: nginx-container
      image: nginx
      volumeMounts:
        - name: shared-data
          mountPath: /usr/share/nginx/html
    - name: alpine-container
      image: alpine
      volumeMounts:
        - name: shared-data
          mountPath: /pod-data
      command: ["/bin/sh"]
      args: ["-c", "echo Hello, World!!! > /pod-data/index.html"]
```
2. EXEC: in resource tree: select Workloads > Pods >`multi-pod2` and pick "Open Terminal" in the context menu
3. ASSERT: terminal to the container `nginx-container` is openened.
4. EXEC: click on `alpine-container` in the list of containers

**Result:**
Fatal IDE error shows up, the details of it you'll get the following:
```
Exec Failure

io.fabric8.kubernetes.client.http.WebSocketHandshakeException
	at io.fabric8.kubernetes.client.okhttp.OkHttpWebSocketImpl$BuilderImpl$1.onFailure(OkHttpWebSocketImpl.java:66)
	at okhttp3.internal.ws.RealWebSocket.failWebSocket(RealWebSocket.java:571)
	at okhttp3.internal.ws.RealWebSocket$2.onResponse(RealWebSocket.java:198)
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:203)
	at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
	Suppressed: java.lang.Throwable: waiting here
		at io.fabric8.kubernetes.client.utils.Utils.waitUntilReady(Utils.java:169)
		at io.fabric8.kubernetes.client.utils.Utils.waitUntilReadyOrFail(Utils.java:180)
[...]

PR #512, opened 2022-11-07
impl'd terminal&log sessions independent of context (#511)

### LINKED ISSUE #511: Terminal: should stay operational even if I switch the current namespace
**Steps:**
1. EXEC: have a terminal to a pod
2. EXEC: switch the current namespace

**Result:**
Terminal freezes

adietish:
fixes #511


adietish:
There's 1 usecase that doesn't work yet:

**Steps:**
1. ASSERT: have 2 contexts: 
  * Red Hat sandbox where your token is expired
  * minikube
2. ASSERT: have a pod running on minikube
3. ASSERT: have minikube as current context
4. EXEC: open a terminal to the pod in minikube
5. EXEC: switch the current context to the Red Hat sandbox

**Result:**
Terminal is freezing. You have to close and open a fresh terminal. The exception that is thrown is 
`java.net.SocketTimeoutException: sent ping but didn't receive pong within 30000ms (after 0 successful ping/pongs)`



If you have a valid token to the sandbox everything works fine. The terminal stays functional.

adietish:
Odd, I now cannot reproduce it any more.

The current implementation should not process anything when the change in the kubeconfig file happens: 
When setting the users set the current context, the plugin updates the kubeconfig file. The update happens async in a thread:
AllContexts
```kotlin
	private fun recreateClient(namespace: String?, context: String?, client: ClientAdapter<out KubernetesClient>?)
			: ClientAdapter<out KubernetesClient> {
		client?.close()
		val new = clientFactory.invoke(namespace, context)
---->		new.config.save()
		this.client.set(new)
		return new
	}
```

The change in file then invokes `AllContexts.onKubeConfigChanged`. This method, that is always invoked when the kubeconfig file changes, then should not do anything because it sees that the kubeconfig file and the client config match in 
* current context
* all contexts
* token

AllContexts
```kotlin
	protected open fun onKubeConfigChanged(fileConfig: io.fabric8.kubernetes.api.model.Config) {
		val client = client.get() ?: return
		val clientConfig = client.config.configuration
--->		if (ConfigHelper.areEqual(fileConfig, clientConfig)) {
			return
		}
		client.close()
		this.client.reset() // create new client when accessed
		refresh()
	}
```

adietish:
Fixed the following usecase: 
**Steps:**
1. EXEC: have a pod running & a terminal for it
2. EXEC: delete the pod
**Result (before)**:
Terminal is freezing as soon as the pod disappears

**Result (now):**
An error in the terminal tab is shown
image

adietish:
I dont see how I can get informed of the connection being closed for logs. For terminals I can provide a listener that gets called whenever the connection is closed. This doesn't exist for logs.

adietish:
@jeffmaury, @lstocchi: this is ready for testing. #502 still exists and may get fixed automatically when using client v6. I'll fix/add unit tests while you're testing this.

lstocchi (approved):
LGTM, as you said i didn't consider this https://github.com/redhat-developer/intellij-kubernetes/issues/502

lstocchi:
A sec maybe i found something when switching namespace on same cluster

lstocchi (changes requested):
Sorry for coming up with an issue after giving my LGTM.... I did my tests by using two different clusters but then it came to my mind i didn't test it by switching namespaces on the same cluster.

It looks to me that when i switch namespace the terminal&log session are still bind to the old ns.
What i do:
1) i'm connected to the default ns on my minikube cluster
2) i open the terminal and logs tabs of one pod
3) everything works fine
4) i switch namespace
5) i try to open the terminal and logs tab of a pod that only exists in this namespace (and not in default - be sure because you may have a resource with the same name on the default ns and the behavior may seem correct)
6) the terminal tab shows the error `it cannot connect` and the log tab a message like `"pods \"nginx-8\" not found"` but the pod exists and it's reachable 
7) i have the impression that it still looks at the default namespace even if i switched

adietish:
@lstocchi Thanks for spotting this. I found and fixed the problem. Please re-test.

lstocchi (approved):

PR #514, opened 2022-11-14
allow to set new context if it is currently null (#513)

### LINKED ISSUE #513: Cannot set new context if there's no current context yet
**Steps:**
1. ASSERT: have no current context set in kubeconfig (to unset your current context run `kubectl config unset current-context`)
2. ASSERT: in resources tree, there are 1+ contexts but there's none that can be expanded
3. EXEC: select a context and pick "Set as Current Cluster" in the context menu

**Result:**
Nothing happens. The selected context is not set as current cluster.

adietish:
fixes #513


lstocchi (approved):
LGTM

PR #518, opened 2022-11-17
use 8-thread (was: 1) invoker when displaying tree items (#493)

### LINKED ISSUE #493: IDE stuck for some second when opening yaml files and not being connected to any cluster
The use case is this. I'm not connected to any cluster and i want to work with yaml-k8s-resource files. 

Everytime i try to open a file, the IDE gets blocked for 3-5 secs because the plugin tries to connect to the cluster to show the notification bar and it failed. Is there anything it could be done to prevent the IDE to be blocked?

Below a gif, just to show the amount of time it gets stuck, which is meaningful. I can click the tree node, manually click start on the timer and it reaches 2-5 secs easily.

kubernetes-loading

adietish:
fixed #493


adietish:
@lstocchi with this change I cannot reproduce the issue any more. My usecase was different than yours though, could you please verify that your case is solved, too?

lstocchi (approved):
LGTM, tried and it works fine now

PR #560, opened 2023-02-09
only watch the given resource, not all in same ns (#559)

### LINKED ISSUE #559: Editor: notifies me that my "old" (before renaming it) resource was deleted
**Steps:**
1. EXEC: open the following yaml in an editor and push it to the cluster:
```yaml
---
apiVersion: "v1"
kind: "Pod"
metadata:
  name: "alpine1"
spec:
  containers:
  - image: "alpine"
    name: "alpine"
    ports:
    - containerPort: 8080
  securityContext:
    fsGroup: 2000
    runAsGroup: 3000
    runAsUser: 1000
```
2. EXEC: change the name to `alpine2` and push it to the cluster
3. EXEC: in the tree: delete `alpine1`

**Result:**
Editor notifies you that `alpine1` was deleted. It should not since I am now editing `alpine2`

This is caused by the following that was erroneously migrated in #500:
`NonCachingSingleResourceOperator` is not specifying the name of the resource that should be used. It is thus watching all resources of the same kind in the namespace of the given resource:
```kotlin
fun watch(resource: HasMetadata, watcher: Watcher<HasMetadata>): Watch? {
        ...
        val op = createOperation(genericKubernetesResource)
        ...
        op.watch(GenericKubernetesResourceWatcherAdapter(watcher))
        ...
```

adietish:
fixes #559

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/resource/NonCachingSingleResourceOperator.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/resource/NonCachingSingleResourceOperator.kt
@@ -51,15 +51,14 @@ class NonCachingSingleResourceOperator(
      * @return resource that was retrieved from cluster
      */
     fun get(resource: HasMetadata): HasMetadata? {
-        return if (hasName(resource)) {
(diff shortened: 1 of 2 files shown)


robstryker:
I am unable to replicate this bug when testing with this pull request.  +1 lgtm

PR #574, opened 2023-02-17
dont display notification outside #update, set status instead & only show push notif (#570)

### LINKED ISSUE #570: Should reduce number of editor notifications, they are detrimental for user focus when editing
The editor currently notifies the user of many things:
* when editor is modified and can be pushed to the cluster (It'll tell you if you create a new resource or modify an existing one.)
* when cluster resource is modified and can be pulled into the editor
* when editor was successfully pushed to the cluster
* when cluster resource was successfully pulled to the editor

The general feeling is that there are too many notifications. 

Ex. when you push the editor to the cluster, the cluster adds managed fields to the resource. In the editor you therefore get a first notification "Push to create/modify" when editing. When you push you then immediately get a second notification that you can pull ("Modified on cluster, Pull?"). If you're editing ex. a pod, the pod changes its status ("created", "started", then "running") and every time this happens you get a notification.

The idea is to not notify "can pull", "successfully pushed", "successfully pulled", but only "can push" if the editor has changes that you can push to the cluster.

The resource states that are not notified any longer should then later get unveiled with inlay hints in the editor

adietish:
fixes #570


adietish:
This requires an overhaul of the editor to allow to decide which of the competing notifications should be displayed (ex. a notification that informs of pushing to the cluster successfully and the notification that changes happened on the cluster and can be pulled).

adietish:
This requires an overhaul of the editor to allow to decide which of the competing notifications should be displayed (ex. a notification that informs of pushing to the cluster successfully and the notification that changes happened on the cluster and can be pulled).

JessicaJHee:
I've played around with this change and reducing the number of notifications is definitely an improvement! 

In the future we may want to consider changing to an even less invasive method to let the user know that there is some un-pushed changes (such as adding `*` beside file name or some indicator on the tool bar icons). But for now this change works great :+1:

PR #577, opened 2023-03-08
corrected IC-2023.1 incompatibilities (#576)

### LINKED ISSUE #576: make compatible with IC-2023.1
image

adietish:
fixes #576


adietish:
This fails when trying to build against `P-ideaVersion=IC-2022.1` with the following error:

```
e: /Users/andredietisheim/Documents/jboss-workspaces/intellij-kubernetes/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/console/ConsoleTab.kt: (213, 13): Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option
```

This is inexplicable to me since the build explicitly states to target jdk11 (`jvmTarget = "11"`)

adietish:
The above compilation error is a bug in the IDE/gradle: https://youtrack.jetbrains.com/issue/KTIJ-20816/Bogus-error-Cannot-inline-bytecode-built-with-JVM-target-11-into-bytecode-that-is-being-built-with-JVM-target-1.8.

sbouchet (approved):
LGTM. installed snapshot in 2023.1 Beta .

PR #596, opened 2023-05-05
Fix badge error on readme

mohitsuman:


mohitsuman:
Close in releation with #596

Issue #602, opened 2023-05-18
Error notification doesnt go away when editing after a push error


adietish:
**Steps:**
1. ASSERT: make sure that you are using kubernetes >= 1.25 (required because pushing to the cluster fails because of schema violation)
1. EXEC: File > New YML file & paste the following resource without name
```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: crontabs.stable.example.com
spec:
  group: stable.example.com
  versions:
    - name: v1
      served: true
      storage: true
  scope: Namespaced
  names:
    plural: crontabs
    singular: crontab
    kind: CronTab
    shortNames:
      - ct
  validation:
    openAPIV3Schema:
      type: object
      properties:
        spec:
          type: object
          properties:
            cronSpec:
              type: string
            image:
              type: string
            replicas:
              type: integer
```
1. ASSERT: Push notification appears telling you that you can create a new resource
1. EXEC: Push
1. ASSERT: error notification ("Could not push...to cluster", Details: "...spec.validation: not declared) is shown
1. EXEC: paste the following corrected resource into the editor:
```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: crontabs.stable.example.com
spec:
  group: stable.example.com
  versions:
    - name: v1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                cronSpec:
                  type: string
                image:
                  type: string
                replicas:
                  type: integer
  scope: Namespaced
  names:
    plural: crontabs
    singular: crontab
    kind: CronTab
    shortNames:
      - ct
```
**Result:**
Error notification doesnt disappear.

**Expected result:**
Error notification should disappear as soon as you modify the editor

PR #604, opened 2023-05-23
corrected erroneous ClusterResource#isSupported() (#602)

adietish:
This error was introduced when fixing #602 and went undetected when IT tested


sbouchet (approved):
LGTM

Issue #600, opened 2023-05-16
"unable to find valid certification path to requested target" when trying to edit a resource


adietish:
This was reported in https://github.com/redhat-developer/intellij-kubernetes/issues/521#issuecomment-1549309067
**Steps:**
1. ASSERT: have an openshift cluster running on a private certificate
2. ASSERT: make sure it's the current context 
3. EXEC: in the plugin, open a resource (double click or "Edit.." in the ctx menu

**Result:**
Editor is opened but errors with the following details
```
PKIX path building failed: usn.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

Adding the certificate under Settings -> Tools -> Server Certificates and also enabling "Accept non-trusted certificates automatically" doesn't help.

To reproduce this you can use our cluster bot:

**Steps:**
1. EXEC: using the clusterbot, launch a new instance (takes ~30min)
2. EXEC: log into the new cluster, copy the login command, paste it into bash/zsh and execute it
```
oc login --token=sha256~XXXXXXX --server=https://XXXXXXXX-east-1.amazonaws.com:6443
```
3. ASSERT: the command errors with the following that confirms that a private certificate is used:
```bash
error: tls: failed to verify certificate: x509: “kubernetes” certificate is not trusted
```
4. EXEC: add ` --insecure-skip-tls-verify=true` to the command
5. ASSERT: login is successful
6. EXEC: edit the entry for the new cluster in ~/.kube/config: remove ` insecure-skip-tls-verify: true`
7. EXEC: launch plugin (and maybe expand the `Projects` node)

**Result:**
The cluster is listed but not detected as OpenShift. Furthermore you cannot list any resources (expand nodes in the tree). They all error ("Error: Could not get ...")
In the IDEA log you discover the error:
```bash
Caused by: java.io.IOException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

adietish:
IDEA has settings for ssl certificates where the user can disable verification and add it's own certificates:
<img width="986" alt="image" src="https://github.com/redhat-developer/intellij-kubernetes/assets/25126/2775019b-cd07-4a58-91b6-ac031283ec4c">
The IDE would also ask the user for confirmation if certificates are not trusted.
[...]

PR #607, opened 2023-05-25
use IDEA setting 'accept non-trusted certificates' (#600)

adietish:


adietish:
Current simplistic implementation is using the IDEA setting "Accept non-trusted certificates automatically". It is very limited though: 
1. The plugin cannot respect a change in this setting without the IDE being restarted. The reason for this is that the IDE wont notify the plugin of a change in this setting.
2. The plugin cannot take the certificates into account, that you manually add in the settings.

A solution to both these limitation would be to hand the IDEA certificate manager over to our kubernetes-client. This isn't possible currently because the client expect the certificate-manager and the key-manager. I need to get the ones used by the client, hand it over to IDEA and give the client the IDEA managers. The client currently wont allow me to access those, it looks as if I need a change in kubernetes-client for this.

adietish:
If you follow the steps with our cluster bot, to reproduce this the plugin now behaves as follows:

You are prompted to accept or reject the untrusted certificate for your cluster:
image
Once you accept it you can deal with your cluster in IDEA

adietish:
Hi @datho7561, can you please test this? You can find the steps to reproduce the issue and test this PR in https://github.com/redhat-developer/intellij-kubernetes/issues/600#issue-1712000083. The variant 2, where you use the slack "Cluster Bot" is the simpler approach. There's no need to install a self-signed certificate since the cluster that the bot creates has a self-signed certificate already. Thanks!

datho7561 (approved):
Looks good, and it's working well for me!

adietish:
Added unit-tests.

PR #624, opened 2023-06-30
cache state, only recreate when new resource is set (#613)

### LINKED ISSUE #613: When using sandbox the editor is freezing for a few seconds on startup
**Steps:**
1. EXEC: have Red Hat Sandbox as current cluster
1. EXEC: create the quarkus kubernetes-quickstart app
```bash
quarkus create app org.acme:kubernetes-quickstart \
    --extension='resteasy-reactive,kubernetes,jib'
```
2. EXEC: run `./mvnw clean package`
3. EXEC: import it into intellij
4. EXEC: open the k8 resource file `target/kubernetes/openshift.yml` in an editor
5. EXEC: scroll around until the editor is freezing

**Result:**
At some point the editor is freezing for a few seconds (maybe 4s) before it displays the notification

adietish:
fixes #613

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/EditorResource.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/EditorResource.kt
@@ -62,6 +62,7 @@ open class EditorResource(
                 && !areEqual(new, existing)) {
                 this.resource = new
             }
+            setState(null) // reset state
         }
     }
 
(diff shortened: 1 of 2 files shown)


adietish:
Creating the `EditorResourceState` the state can reach out to the cluster to check if resource is outdated among other tests (the state can be `Modified`, `Outdated`, `DeletedOnCluster`, `Pushed`, `Pulled`, `Error`) etc. We therefore should only create the state when the `EditorResource` is created or when the `EditorResource` gets a new resource. Any other time the state is retrieved (`EditorResource#getState`), a cached state should be returned.

sbouchet (approved):
LGTM. no more freeze on the editor when using devsandbox.

PR #626, opened 2023-07-04
set ideaVersion to latest to dev against latest (#625)

### LINKED ISSUE #625: manually set since-build in plugin.xml, bump IC version in gradle.properties to latest
`<idea-version since-build="XXX"/>` in `plugin.xml` sets the minimal version required by the plugin. `ideaVersion` in `gradle.properties` sets the version used when developing. We should manually set both. We currently only set the version in `gradle.properties` and have it copied over to `plugin.xml` via intellij gradle plugin.

adietish:
fixes #625


sbouchet (approved):
LGTM

PR #627, opened 2023-07-08
Issue 617

adietish:


datho7561 (approved):
Looks good!

I tried it out with:
- crc OpenShift
- minikube
- crc microshift

OpenShift and minikube worked well. MicroShift doesn't come with a dashboard, so when I tried it out, there was an error popup, as expected.

PR #635, opened 2023-07-20
only reset resource state if resource is modified (#634)

### LINKED ISSUE #634: If I push to a cluster and it errors, I am not notified of the error but prompted to push again
**Steps:**
1. EXEC: open an editor for the following resource:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-1
  namespace: bogus
spec:
  containers:
    - name: nginx
      image: nginx
      ports:
        - containerPort: 80
```
2. ASSERT: verify that the `namespace` points to a namespace that doesn't exist on your cluster
3. EXEC: push to the cluster

**Result:**
You are prompted to push again, right after the push, which made the notification disappear, 

**Result:**
When pushing fails, an error should be displayed. You should not be prompted to push again.

adietish:
fixes #634


datho7561 (approved):
Looks good!

PR #638, opened 2023-07-21
using proper chainging load ns -> display dialog (#637)

### LINKED ISSUE #637: Text field for the current namespace isn't displayed at times
I cannot replicate this (I'm on Mac), it was reported by @olkornii (using Linux) while doing the integration tests. It happens to him maybe 2 out of 3 times

**Steps:**
1. EXEC: In the resource tree: select the "Namespaces" category. Right click on it and pick "Set Current Namespace" (if you are using a kubernetes cluster. If it's OpenShift it's "Set Current Project"

**Result:**
A dialog shows up, that allows you to type the name of the namespace to use as current namespace. At times the dialog shows up without the text field (on Linux, not on Mac. Dont know about Windows)
image

adietish:
possibly fixes #637


JessicaJHee (approved):
Works as expected on my Linux machine, LGTM!

PR #639, opened 2023-07-21
prevent both tool windows from waiting for indexing (#636)

### LINKED ISSUE #636: on startup, resource tree is not available as long as IDEA is indexing
**Steps:**
1. EXEC: launch IDEA with the plugin installed
2. ASSERT: make sure that IDEA is indexing your project
2. EXEC: immediately after the start switch to the resource tree (while indexing still is progressing)

**Result:**
The resource tree is not available, a message tells you that it is not available as long as IDEA is indexing your project
image

adietish:
fixes #636


JessicaJHee (commented):
To confirm, the expected behavior is that on start up while the views are loading, it says `Nothing to show` in the window instead of `This view is not available until the indexes are built`?

JessicaJHee (approved):
LGTM, thanks for the clarification!

PR #641, opened 2023-07-24
correctly synchronize accessing client when config changes (#640)

### LINKED ISSUE #640: Use current Namespace/Project has a bug with minikube
Configurations:
  Kubernetes v1.26.2 on Docker 24.0.4
  openjdk 17.0.7 2023-04-18
  OS: Fedora 38
  Intellij 2023.1

Steps to reproduce:
1. Start minikube cluster
2. Expand Namespaces
3. Select namespace -> right click -> Use namespace -> tree refreshed (WRONG BEHAVIOR)
4. Repeat few times -> cluster name changes for short time (sometimes freezes)

On the video's 10 second is a bug.
[Screencast from 2023-07-24 12-22-10.webm](https://github.com/redhat-developer/intellij-kubernetes/assets/60704619/0c129321-6c3a-46a9-ae84-877bd7eec67d)

adietish:
possibly fixes #640 (not sure because I cannot replicate the issue)

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/AllContexts.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/AllContexts.kt
@@ -143,6 +143,7 @@ open class AllContexts(
 				this.client.get()
 			)
 			this.current?.close()
+			newClient.config.save().join()
 			all.clear() // causes reload of all contexts when accessed afterwards
 			val newCurrent = this.current // gets new current from all
 			if (toWatch != null) {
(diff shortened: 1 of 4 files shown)


datho7561:
This fixes the issue where the context is replaced with <unknown context> in the tree, but the tree view still refreshes

datho7561:
I tried again, but the latest change doesn't prevent the tree view from refreshing

adietish:
@datho7561 thanks for testing. So I'd suggest to merge the PR since it fixes the critical issue where the context would become <unknown context>. I'd then keep the issue open (or create a new one) for the remaining issue where the refresh is still happening.

datho7561 (approved):
Yes, that sounds like a good idea!

PR #651, opened 2023-08-15
list replica sets & replication controllers (#650)

### LINKED ISSUE #650: As a user I want to have Replica Sets  and Replication Controllers listed in the resource tree
**Steps:**
1. EXEC: paste the following into an editor:
```yaml
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: frontend
  labels:
    app: guestbook
    tier: frontend
spec:
  # modify replicas according to your case
  replicas: 3
  selector:
    matchLabels:
      tier: frontend
  template:
    metadata:
      labels:
        tier: frontend
    spec:
      containers:
      - name: php-redis
        image: gcr.io/google_samples/gb-frontend:v3
```
2. EXEC: push editor to the cluster
3. EXEC: inspect resource tree for your new replica set

**Result:**
Replica sets are not listed, there's no category where they would appear.

**Expected result:**
The resource tree should have `[cluster] > Workloads > Replica` Sets and list the above resource there
VSCode is missing them, jetbrains plugin is listing them:
image

Same is true for Replication Controllers:
```yaml
apiVersion: v1
kind: ReplicationController
metadata:
  name: nginx
spec:
  replicas: 3
  selector:
    app: nginx-rc
  template:
    metadata:
      name: nginx-rc
      labels:
        app: nginx-rc
    spec:
      containers:
      - name: nginx-rc
        image: nginx
        ports:
        - containerPort: 80
```
Jetbrains Kubernetes plugin is listing them:
image

adietish:
fixes #650 

both "Replication Controllers" and "Replica Sets" are now listed with their pods:
image


adietish:
@sbouchet awesome, thanks for reviewing.

PR #653, opened 2023-08-21
show error details when fetching children fails (#20)

### LINKED ISSUE #20: Display error if current cluster is not reachable

adietish:
fixes #20


adietish:
Wrong issue, closing.

PR #654, opened 2023-08-21
show error details when fetching children fails (#28)

### LINKED ISSUE #28: Error nodes: improve text to include the reason for the failure
When resources cannot be loaded (bcs of ex. permissions or connection timeout), the error node only reports the failure, doesn't mention the reason. It should also mention why loading failed.

adietish:
fixes #28

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/tree/TreeStructure.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/tree/TreeStructure.kt
@@ -248,8 +248,35 @@ open class TreeStructure(
         project
     ) {
         override fun getLabel(element: java.lang.Exception): String {
-            return "Error: ${element.message ?: "unspecified"}"
+            return "Error: ${getMessage(element)}"
         }
+
+        private fun getMessage(e: Exception): String {
+            val causeMessage = e.cause?.message
+            return if (causeMessage == null) {
+                getMessageOrDefault(e)
+            } else if (causeMessage.contains("Operation: ")) {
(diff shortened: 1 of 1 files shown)


adietish:
With this PR the cause is displayed instead of a generic failure text which tells you that the resources could not have been retrieved.
image

adietish:
To test this one needs a Red Hat Developer Sandbox with an expired token (ex. create the cluster, try to use it after 24h) and a minikube that's defined in kube config but is not running (ex. start minikube, copy kube config, stop minikube, overwrite kube config with the copy). Then:
1. EXEC: on Red Hat Sandbox: expand "Projects" node in resource tree
-> child node displays error telling you that the token is expired (see screenshot above)
2. EXEC: on Red Hat Sandbox: expand "Namespaces" node in resource tree
-> child node displays error telling you that the request is forbidden (see screenshot above)
3. EXEC. on minikube: list any category
-> child node displays an error telling you that the request timed out

adietish:
@datho7561 i am now looking if there's a `DeploymentConfig` for the selected `ReplicationController` and scale that one instead if it exists. Accordingly I need to now tell you what resource is being scale. I print it in the 1st line of the dialog:
image
I did not add anything like "Resource: " or "Scale <resource>" in front of the resource label. That this dialog is about scaling is already visible in the dialog title. Then the kind is printed in front of the name which is - I think - sufficient to grasp that this is the resource that this concerns. Would you agree?

sbouchet (approved):
LGTM. getting real cause messages in tree as expected by this PR

PR #667, opened 2023-11-14
handle null file configs (when removed) when notified of changes

adietish:
This change is required by the change in ConfigWatcher at https://github.com/redhat-developer/intellij-common/pull/202

The current implementation in intellij-common was caused by https://github.com/redhat-developer/intellij-kubernetes/issues/640 and implemented as for https://github.com/redhat-developer/intellij-common/issues/190


datho7561:
If I delete my `~/.kube/config`, nothing happens, even if I manually refresh the tree (all the entries are there). Is this expected?

datho7561:
When I close and reopen the IDE after deleting my `~/.kube/config`, it now shows no clusters as available, until I recreate the `~/.kube/config`, at which point it updates with the cluster I added automatically.

adietish:
@datho7561 That's expected. Avoiding refreshes when the file is null was my attempt to avoid the problems described in #640. I don't know how else to handle the additional notifications on Linux/Windows that you discovered in https://github.com/redhat-developer/intellij-common/issues/190#issuecomment-1650137863. If you see a way I'd be very grateful.

datho7561 (approved):
Alright, if this is the expected behaviour then this seems to be working!

adietish:
@datho7561 it would be nice if it could empty the tree when you remove the kube config, but I dont see how I can achieve that. I am trying to identify partial updates on Linux/Windows (the 3 notifications that you identified) by the fact that the kube config is null. On the other hand, the kube config is null, too when you delete the file. I dont see how I can achieve both. Do you see a way, maybe?

datho7561:
No, sorry, I'd just completely forgot the context for this PR, that's why I was expecting different behaviour.

adietish:
The alternative approach found in https://github.com/datho7561/intellij-common/commit/51fadfe0763fe7def7882ee3e33f49c85337ea4d#diff-7dbd2b293c8d442ed99e82cd50e006eacdabdd7befe34b2f1b7ad3c545c97712R110-R117 
```java
    private Config loadConfig(Path path) {
        if (path == null) {
            return null;
        } else {
            try {
                if (Files.exists(path, new LinkOption[0]) && this.isConfigPath(path) && Files.size(path) > 0L) {
```
was implemented in intellij-common: https://github.com/redhat-developer/intellij-common/blob/fef9547fb1fda4ea6975d0538745197ab9c53052/src/main/java/com/redhat/devtools/intellij/common/utils/ConfigWatcher.java#L115-L117
```java
if (Files.exists(path)
                    && isConfigPath(path)
                    && Files.size(path) > 0) {
```

I thus suggest that we merge this PR as-is and eventually file an issue for the plugin not reacting to the removal of the kube config.

PR #670, opened 2023-11-21
dont set namespace when switching (kubeconfig) context (#669)

### LINKED ISSUE #669: "Current namespace" of my old context is overwritten when I switch to a new context
The following can be applied for any setup with 2 clusters

**Steps:**
1. ASSERT: have 2 contexts with different "namespace"s 
* minikube with namespace named "default"
* redhat sandbox with namespace "<YOUR_USERID>-dev" ("adietish-dev" for me)
```yaml
contexts:
- context:
    cluster: "api-sandbox-m3-1530-p1-openshiftapps-com:6443"
    namespace: "adietish-dev"
    user: "adietish/api-sandbox-m3-1530-p1-openshiftapps-com:6443"
  name: "adietish-dev/api-sandbox-m3-1530-p1-openshiftapps-com:6443/adietish"
- context:
    cluster: "minikube"
    namespace: "default"
    user: "minikube"
  name: "minikube"
```
2. ASSERT: have "minikube" set as current cluster/current namespace
3. EXEC: in tree: select "sandbox" and pick "Set as Current Cluster" in context menu
4. EXEC: open `~/.kube/config` in an editor and verify the "namespace" for "minikube"

**Result:**
"namespace" for "minikube" was set to "<YOUR_USERID>-dev" (was "default")
```yaml
- context:
    cluster: "minikube"
    namespace: "adietish-dev"
    user: "minikube"
  name: "minikube"
```

adietish:
fixes #669

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/client/ClientConfig.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/client/ClientConfig.kt
@@ -54,15 +54,14 @@ open class ClientConfig(private val client: Client, private val executor: Execut
 					return@supplyAsync false
 				}
 				val fromFile = kubeConfig.load() ?: return@supplyAsync false
-				val currentContextInFile = KubeConfigUtils.getCurrentContext(fromFile)
 				if (setCurrentContext(
 						currentContext,
-						currentContextInFile,
+						KubeConfigUtils.getCurrentContext(fromFile),
 						fromFile
 					).or( // no short-circuit
 						setCurrentNamespace(
 							currentContext?.context,
-							currentContextInFile?.context
+							KubeConfigUtils.getCurrentContext(fromFile)?.context
 						)
 					)
 				) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/client/ClientConfig.kt
```diff
@@ -54,15 +54,14 @@ open class ClientConfig(private val client: Client, private val executor: Execut
 					return@supplyAsync false
 				}
 				val fromFile = kubeConfig.load() ?: return@supplyAsync false
-				val currentContextInFile = KubeConfigUtils.getCurrentContext(fromFile)
```
adietish, line 57: This was the error: storing the current context before the new context is set and then setting the namespace on it:
`#setCurrentContext` would change the current context and `#setCurrentNamespace` would then change the namespace on the current context that was "current" before changing to a new one.

datho7561 (approved):
This change works for me

PR #674, opened 2023-11-30
use ClusterHelper#isOpenShift in intellij-common

adietish:
related to https://github.com/redhat-developer/intellij-common/pull/203


src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/client/ClientAdapter.kt
```diff
@@ -102,11 +102,7 @@ abstract class ClientAdapter<C : KubernetesClient>(private val fabric8Client: C)
         }
 
         private fun isOpenShift(osClient: NamespacedOpenShiftClient): Boolean {
```
sbouchet, line 104: remove this private metho and do a direct call ?

PR #692, opened 2023-12-18
use default/1st namespace if there's no current (#668)

### LINKED ISSUE #668: As a user I can be puzzled if I have no  current namespace
**Steps:**
1. ASSERT: have a 
  * `PersistentVolume` and a 
  * `PersistentVolumeClaim`
2. EXEC: unset the namespace in your current context:
```bash
kubectl config unset contexts.<NAME OF YOUR CURRENT CONTEXT>.namespace
```
3. EXEC: in the resource tree: expand 
  * `[cluser] > Storage > Persistent Volumes` and 
  * `[cluser] > Storage > Persistent Volume Claims`

**Result:**
You only see the persistent volume(s) listed, persistent volume claims are not listed. 

This behaviour is perfectly expected, it's not a bug, but it can be puzzling if you are not aware that you dont have the current namespace set.

We should either make a missing current namespace more prominent (ex. paint `current: <none>`, left of `[cluster] > Namespaces`, red) or list the "default" or 1st namespace instead.

adietish:
fixes #668


sbouchet (approved):
LGTM. 
correctly use default or first found namespace when unsetting current namespace manually.

PR #703, opened 2024-01-09
only show Refresh for active contexts (#702)

### LINKED ISSUE #702: Refresh is available for "inactive" contexts of the resource tree
**Steps:**
1. ASSERT: Have 2+ contexts in your kube config
2. EXEC: right click on an inactive context and inspect the context menu

**Result:**
image
There's an entry "Refresh" for the inactive context. It doesn't make sense to be able to refresh an inactive context.

adietish:
fixes #702


sbouchet (approved):
LGTM

PR #707, opened 2024-01-11
allow to force delete resources (#572)

### LINKED ISSUE #572: Should allow to force delete resource(s)
in kubectl one can force the deletion of ex. a pod:
```
kubectl delete pod nginx --force
```
This allows one to forcefully remove a resource that is stuck in terminating state.
We should allow the same.

adietish:
fixes #572


sbouchet (approved):
LGTM. force delete behave like expected, UI  shows immediate deletion as expected.

PR #709, opened 2024-01-11
check if resource exists BEFORE pushing -> Created state for new resources (was: Updated) (#708)

### LINKED ISSUE #708: When pushing to create a new resource, notification says that resource was updated
**Steps:**
1. ASSERT: make sure you dont have a pod named `alpine1` on your cluster
2. EXEC: open the following in an editor
```yaml
---
apiVersion: "v1"
kind: "Pod"
metadata:
  labels:
    jedi: "luke"
  name: "alpine1"
spec:
  containers:
  - image: "alpine"
    name: "alpine"
    ports:
    - containerPort: 8080
```
3. EXEC: push the editor to create a new pod on the cluster
4. ASSERT: new pod `alpine1` is created
5. EXEC: look at the notification in the editor, that tells you what was done on the cluster

**Result:**
image
You're told that the pod was `Updated`

**Expected result:**
You should be informed that the pod was `Created`

adietish:
fixes #708

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ClusterResource.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/ClusterResource.kt
@@ -84,7 +84,7 @@ open class ClusterResource protected constructor(
                 return
             }
             this.updatedResource = resource
-            setDeleted(false)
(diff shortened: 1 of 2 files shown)


sbouchet (approved):
LGTM. message in toolbar is coherent with what's going on cluster.

PR #714, opened 2024-02-12
feat: show decoded value for base64 values in Secrets, ConfigMaps (#663)

### LINKED ISSUE #663: As a user I would like to easily create/edit Secrets in the editor
3rd most upvoted feature in vscode-kubernetes-tools: https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/issues/579

The idea is to handle base64 encoding transparently like https://github.com/rajatjindal/kubectl-modify-secret does. Secret is presented decoded and is encoded automatically when pushed to the cluster.

adietish:
fixes #663 

`Secrets` and `ConfigMaps` to test this PR: [secrets-configmaps.zip](https://github.com/redhat-developer/intellij-kubernetes/files/14336192/secrets-configmaps.zip)

### DIFF
--- a/.github/workflows/IJ.yml
+++ b/.github/workflows/IJ.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        IJ: [IC-2021.1, IC-2021.2, IC-2021.3, IC-2022.1, IC-2022.2, IC-2022.3, IC-2023.1, IC-2023.2, IC-2023.3]
+        IJ: [IC-2022.1, IC-2022.2, IC-2022.3, IC-2023.1, IC-2023.2, IC-2023.3]
 
     steps:
     - uses: actions/checkout@v2
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/SwingUtils.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/SwingUtils.kt
@@ -1,0 +1,30 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.redhat.devtools.intellij.kubernetes
+
+import com.intellij.ui.components.JBLabel
+import com.intellij.util.ui.JBUI
+import com.intellij.util.ui.UIUtil
+import javax.swing.text.JTextComponent
+
+fun createExplanationLabel(text: String): JBLabel {
+	return JBLabel(text).apply {
+		componentStyle = UIUtil.ComponentStyle.SMALL
+		foreground = JBUI.CurrentTheme.Link.Foreground.DISABLED
+	}
+}
+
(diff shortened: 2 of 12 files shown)


adietish:
Current impl works fine for 1-line secrets:
Screen-Recording-2024-02-15-at-1

I am current working on displaying hints and editing multiline secrets.

adietish:
I now added display & edit for multiline values:
image

adietish:
I decided to have binary data shown as hex-values, not allowing them to be edited. 

image

I could have provided them to be edited in binary form via a `JFormattedTextField` (see https://docs.oracle.com/javase/tutorial/uiswing/components/formattedtextfield.html) but I'm unsure if this would be useful and match real-world use cases. I thus didn't go for it for now.
One could have a f

adietish:
Here are some examples that can be used to test this PR: [secrets-configmaps.zip](https://github.com/redhat-developer/intellij-kubernetes/files/14336192/secrets-configmaps.zip)

adietish:
@sbouchet, @datho7561: please review. I attached a zip with several secrets and configmaps to test this. Thanks :)

adietish:
While waiting for the reivew I'm adding unit-tests.

datho7561:
I connected to OpenShift Sandbox and opened the secrets files for the service accounts, but none of the values are show as inlay hints. I expect this is because all the values are all in double quotes.

datho7561:
When working with a string with newlines in it, if I want to add another newline, I can add the first one, but when I try to add a second one, it closes the text box and applies the new value.

adietish:
> I connected to OpenShift Sandbox and opened the secrets files for the service accounts, but none of the values are show as inlay hints. I expect this is because all the values are all in double quotes.

thanks for finding this, I'll look into it and try to fix it.

adietish:
> When working with a string with newlines in it, if I want to add another newline, I can add the first one, but when I try to add a second one, it closes the text box and applies the new value.

@datho7561: Me thinks that it would be only possible to allow several empty lines when confirming the dialog would be via a dedicated button. Worth it?

adietish:
> > Worth it?
> 
> I think so

Ok, adding a button then. Thanks for pointing this out 🙂

sbouchet:
same comments as @datho7561 , display is correct, edition is ok for single lines values and need chenges for multiline

sbouchet:
latest changes did not changed behavior for multiple lines edit (second hit on ENTER closes the dialog).

it's even worst because now multiple line aren't displayed on local yaml
Capture d’écran du 2024-02-21 14-57-13

but when pulling cluster resources it's ok
Capture d’écran du 2024-02-21 14-57-28

sbouchet:
still does not work as expected on fedora
[Capture vidéo du 2024-02-22 09-56-44.webm](https://github.com/redhat-developer/intellij-kubernetes/assets/600414/15a5c328-4ba2-4581-8af0-a510326a2b04)

datho7561:
Multiline editing now works properly for me.

One more thing I ran into was if you have a single line secret, and try to insert a newline, then it won't work. However, I don't think this is a big issue.

adietish:
@datho7561, @sbouchet: I fixed both issues. One can now edit quoted values and you can insert new lines by pressing Shift + Enter. An explanatory label to the left of the balloon would tell you about it. Enter alone applies the value. I dont provide different views (text field / text area) any more. I always use a text area with at least 3 rows. I noticed that this distinction doesnt make a lot of sense when you start turning a short value to a multi-line value (ex 42 -> big certificate).
I also enabled word wrapping so that the input balloon always has the same width.
image

datho7561 (approved):
Looks great to me!

datho7561 (approved):
Still seems good to me!

PR #733, opened 2024-03-28
fix: display evident error msg instead of full details if unauthorized (#732)

### LINKED ISSUE #732: Tree: Error when unauthorized is cryptic, should be more evident
image

adietish:
fixes #732 

image


sbouchet (approved):
LGTM

PR #740, opened 2024-04-04
fix: corrected status bar msg 'Switching namespace' (#739)

### LINKED ISSUE #739: Switching namespaces displays object in status message
**Steps:**
1. EXEC: select namespace in [cluster] > Namespaces and pick "Use Namespace"
1. EXEC: look at the message in the status bar

**Result:**
The message says "Switching namespace Namespace(apiVersion=v1....". It should say "Switching to namespace [namespace-name]" instead

adietish:
fixes #739


sbouchet (approved):
LGTM

PR #750, opened 2024-04-17
fix: make build pass marketplace verification (#749)

### LINKED ISSUE #749: 1.3.0 is incompatible to all IC starting with IC-2022.3  up to IC-2024.1
related to https://github.com/redhat-developer/intellij-common/issues/222

image

The problem here is `com.redhat.devtools.intellij.common.utils.function.TriConsumer` which lives in intellij-common 1.9.4. Without it, there's only an incompatiblity to `IC-2022.3` left:

image

adietish:
fixes #749


sbouchet (approved):
LGTM

Issue #746, opened 2024-04-15
IT as consistently failing since sinceIdea was bumped to IC-2024.1


adietish:
```java
* What went wrong:
    com.intellij.remoterobot.utils.WaitForConditionTimeoutException: Exceeded timeout (PT2S) for condition function (Failed to find 'Button' by '//div[@text='Kubernetes']' in 2s) 
        at com.intellij.remoterobot.utils.RepeatUtilsKt.waitFor(RepeatUtils.kt:50)
        at com.intellij.remoterobot.utils.RepeatUtilsKt.waitFor$default(RepeatUtils.kt:36)
        at com.intellij.remoterobot.SearchContext$find$1.invoke(SearchContext.kt:53)
        at com.intellij.remoterobot.SearchContext$find$1.invoke(SearchContext.kt:51)
        at com.intellij.remoterobot.stepsProcessing.StepWorkerKt.step(StepWorker.kt:23)
        at com.intellij.remoterobot.SearchContext$DefaultImpls.find(SearchContext.kt:51)
        at com.intellij.remoterobot.fixtures.ContainerFixture.find(ContainerFixture.kt:12)
        at com.intellij.remoterobot.fixtures.CommonContainerFixture.button(CommonContainerFixture.kt:248)
        at com.redhat.devtools.intellij.commonuitest.fixtures.mainidewindow.toolwindowspane.AbstractToolWinPane.stripeButton(AbstractToolWinPane.java:107)
        at org.jboss.tools.intellij.kubernetes.BasicTests.openKubernetesTab(BasicTests.java:120)
        at org.jboss.tools.intellij.kubernetes.BasicTests.connect(BasicTests.java:62)
```

olkornii:
Closing issue. Last ITs are green.

PR #753, opened 2024-04-22
fix: Bumped test library version and intellij version for itests

olkornii:
Bumped intellij-common-ui-test library and intellij version for integration tests.
Disabled createResourceByEdit test. Requires rework because of new minikube version.


olkornii:
Fixes #746

adietish (approved):
looks good. LGTM

PR #758, opened 2024-04-30
fix: removed requested kubernetes version that's too low for github

adietish:
ITs are not fun any more with the following error:
```bash
. ./scripts/build_local.sh
* minikube v1.33.0 on Ubuntu 20.04
! Specified Kubernetes version 1.16.4 is less than the oldest supported version: v1.20.0. Use `minikube config defaults kubernetes-version` for details.
! You can force an unsupported Kubernetes version via the --force flag
```

### DIFF
--- a/scripts/build_local.sh
+++ b/scripts/build_local.sh
@@ -3,7 +3,7 @@
 # Note: run from root
 # This is used to run and connect minikube cluster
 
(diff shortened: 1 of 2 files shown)


olkornii (approved):
LGTM for me. @adietish check if my commit is ok as well please :)

adietish:
@olkornii thanks for helping to fix the ITs 😄

PR #761, opened 2024-05-14
test: Fixed integration tests

### LINKED ISSUE #754: Rework "createResourceByEdit" test
Caused by #746 
The new version of minikube does not allow us to create a new node based on the default node ("minikube").
A new simple test node must be created before running all tests.

olkornii:
Fixes #754 and #746


adietish (approved):
awesome, thanks! LGTM

PR #766, opened 2024-07-02
chore: Update release to ask release version to be published

sbouchet:


.github/workflows/release.yml
```diff
@@ -100,23 +108,24 @@ jobs:
         id: changelog
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          current-tag: ${{env.PLUGIN_VERSION}}
+          current-tag: ${{ inputs.release_version }}
```
adietish, line 111: shouldnt the tag be matching the version that is set to `gradle.properties`?

.github/workflows/release.yml
```diff
@@ -72,10 +80,10 @@ jobs:
           if git diff --quiet; then
             echo "No changes to commit."
           else
-            git commit -sam "chore(skip-release): set version to $PLUGIN_VERSION"
+            git commit -sam "chore(skip-release): set version to ${{ inputs.release_version }}"
           fi
-          git tag $PLUGIN_VERSION
-          git push origin $PLUGIN_VERSION
+          git tag ${{ inputs.release_version }}
```
adietish, line 85: shouldnt the tag be matching the version that is set to `gradle.properties`?

I'm not sure that's why I am asking: I see `NEW_VERSION` being set to `${{ inputs.release_version }}.${{ github.run_number }}` and then being written to `gradle.properties`. But then when tagging only `inputs.release_version` is used. What am I mIssing?

.github/workflows/release.yml
sbouchet: yes, it is the same version, we set it at the very first steps.

adietish (approved):
LGTM

PR #769, opened 2024-07-09
fix: use PopupHandler#installPopupMenu, remove adapter

adietish:
We dont need the `PopupHandlerAdapter` any more since we're using a min version > `IC-2022.1`. Thus removing it and replacing by the API available to >= `IC-2022.1`


sbouchet (approved):
LGTM

PR #790, opened 2024-10-02
feat: allow to disable editor notifications (#763)

### LINKED ISSUE #763: Option to disable PushNotification
The PushNotification is not useful for GitOps workflows, and it is distracting to the user. There should be an option to disable the PushNotification. (i.e. the "Push to create Deployment 'foobar'?" notification)

adietish:
fixes #763

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/EditorResource.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/editor/EditorResource.kt
@@ -33,7 +33,10 @@ open class EditorResource(
     private val resourceModel: IResourceModel,
     private val resourceChangedListener: IResourceModelListener?,
     // for mocking purposes
-    private val clusterResourceFactory: (resource: HasMetadata, context: IActiveContext<out HasMetadata, out KubernetesClient>?) -> ClusterResource? =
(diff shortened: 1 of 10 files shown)


adietish:
@sbouchet:
how to test this:
https://github.com/redhat-developer/intellij-kubernetes/issues/763#issuecomment-2391607391 is describing what's provided in this enhancement/feature.

My first question is: is the label for the checkbox explaining what happens if you check/uncheck it? 

The most basic is test could look like this:

**steps:**
1. ASSERT: make sure that editor sync is enabled in the settings
2. EXEC: create a new yaml file and copy & paste the following into it:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: base64encoded
data:
  foo: text value that noone ca
binaryData:
  bar: jmjKxUGNs/WqDQ==
```
3. ASSERT: editor displays a notification telling you that you can push the editor to the cluster
4. EXEC: go to settings and uncheck "Sync editor with cluster", then Apply/OK
5. ASSERT: notification disappears
6. EXEC: hit the push icon in the toolbar 

**Result:**
The config map in the editor was created on the cluster. You are told by a notification that informs you that the config map was created. In the resource tree you now have a new config map named "base64encoded"

sbouchet (approved):
LGTM. good work !

PR #797, opened 2024-10-09
fix: corrected warnings

adietish:


sbouchet (approved):
LGTM. thanks for quick resolution !

PR #801, opened 2024-10-14
chore: migrate build to 2024.2

sbouchet:
fixes #780

Issue #779, opened 2024-08-02
Cannot get my clusters to appear


creckord:
I have been using the Jetbrains Kubernetes plugin on and off for a while now, and have several clusters added in IntelliJ.

I wanted to give the Redhat Kubernetes plugin a shot, because really, the Jetbrains one is too clunky as my main k8s console (coming from OpenLens).

Unfortunately, all I got from the Redhat plugin was a long time of cluster resource refreshing, but at the end, the Kubernetes tool window just stayed empty. No errors in the logs either.

IntelliJ IDEA 2024.1.4 (Ultimate Edition) Build #IU-241.18034.62
Redhat Kubernetes Plugin 1.3.0
OpenJDK 64-Bit Server VM by JetBrains s.r.o. Runtime version: 17.0.11+1-b1207.24 amd64

creckord:
Hm, actually, looking at a fresh log from a recently restarted IDEA, I get a whole bunch of these from the IntelliJ Kubernetes provider:

```
2024-08-02 20:43:38,680 [   5436]   INFO - #c.i.k.a.KubernetesApiProvider - Configuration refresh started
2024-08-02 20:43:38,951 [   5707]   WARN - #c.i.k.a.KubernetesApiProvider - Configuration not loaded: Context 'prod-gcp-star' not found in contexts list [], kubeconfig file: file://C:/Users/creckord/.kube/config
[...]

PR #806, opened 2024-10-25
revert: "fix: allow multiple configs listed in KUBECONFIG (#779)"

adietish:
reverts #779, commit d5972d2a7068894c4ac06348e8be923e6c9d7d4a.

We merged #779 because things compiled and ran nicely. But then, when implementing the means to compile against `IC-2024.2` plugin verifier started to complain (see #801) because it didnt see the modified kubernetes-client classes that existed in the plugin, that were shadowing the ones in the client-jar.

I tried the following without success:
* moved the classes `src/main/java` -> `src/main/kotlin` (verifier still complained)
* renamed the package with the shadowing classes (invovles copying even more classes, changing classes in the plugin)
* masked the calls to new method that dont exist in the client-jar using reflection (requires changing classes in the plugin)

When trying reflection we decided to abandon the stop gap solution and wait for the release of the kubernetes-client 7.0.0 which should include the required changes.


sbouchet (approved):
Approved, let's hope this will un-block #801 and let's also wait for kubeclient 7.0.0 to be out

Issue #809, opened 2024-12-20
if the kubeconfig is invalid, the tool window stays empty


adietish:
**Steps:**
1. ASSERT: have an invalid kube config 
(ex. minikube port which is invalid/out of bounds `server: "https://192.168.105.4:84437777"`, the correct port is `8443`)
2. EXEC: launch plugin

**Result:**
The toolkit window stays empty.
<img width="338" alt="Image" src="https://github.com/user-attachments/assets/b84a9917-d731-4744-aa00-f2a859d3c1a8" />

adietish:
The reason is that the client that is created throws an exception. The exception is swallowed, no context is created. It should create an "error context" instead.
[AllContexts.kt:](https://github.com/redhat-developer/intellij-kubernetes/blob/719d72c7259d85875e2bc96010d96a3d846a6eab/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/AllContexts.kt#L78)
```kotlin
	override val all: List<IContext>
		get() {
			lock.write {
				if (_all.isEmpty()) {
					try {
						val all = createContexts(client.get(), client.get()?.config)
						_all.addAll(all)
					} catch (e: Exception) {
						//
					}
				}
				return _all
			}
		}
```

adietish:
The refresh that's available in the context menu has no effect because there's no selected node.
https://github.com/redhat-developer/intellij-kubernetes/blob/719d72c7259d85875e2bc96010d96a3d846a6eab/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/actions/RefreshAction.kt#L26

The refresh action should not return without doing anything. It should call `ResourceModel.invalidate()` instead, which then should refresh the whole tree.
https://github.com/redhat-developer/intellij-kubernetes/blob/719d72c7259d85875e2bc96010d96a3d846a6eab/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceModel.kt#L157-L163

adietish:
Furthermore, there's no healing. Correcting the config file wont heal the toolkit window. It stays empty.

PR #810, opened 2024-12-20
fix: display kubeconfig parse error on startup (#809)

adietish:
fixes 
* #809
* #813

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 # libraries
 kubernetes-client = "7.0.0"
-devtools-common = "1.9.7-SNAPSHOT"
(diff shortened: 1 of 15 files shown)


adietish:
@sbouchet: to test this:

**Steps:**
1. ASSERT: have an invalid kube config
    (ex. minikube port which is invalid/out-of-bounds: "https://192.168.105.4:84437777", the correct port is 8443)
1. EXEC: launch plugin

**Result:**
The toolkit window stays empty.

**Next step:**
1. EXEX: correct the port

**Result:**
The tree heals: the error is replaced by the contexts.

Thanks for reviewing this!

sbouchet (approved):
nice improvement ! LGTM.

PR #811, opened 2025-01-21
fix: allow to refresh all contexts (#809)

adietish:
The refresh that's available in the context menu has no effect if no context is selected.
https://github.com/redhat-developer/intellij-kubernetes/blob/719d72c7259d85875e2bc96010d96a3d846a6eab/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/actions/RefreshAction.kt#L26

The refresh action should not return without doing anything. It should call `ResourceModel.invalidate()` instead, which then should refresh the whole tree.
https://github.com/redhat-developer/intellij-kubernetes/blob/719d72c7259d85875e2bc96010d96a3d846a6eab/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/model/ResourceModel.kt#L174-L176


sbouchet (approved):
LGTM. refresh action seems to refresh all context if nothing selected.

PR #816, opened 2025-01-22
fix: dont show actions for selected items when there's no selection (#812)

### LINKED ISSUE #812: "Use Namespace" ctx menu item is visible even though no namespace is selected
**Steps:**
1. ASSERT: no item is selected in the resource tree. You can unselect items by Ctrl + click
2. EXEC: right click in the void area in the resource tree (so that you dont select an item) and open the context menu

**Result:**

Image
"Use Namespace", "Use Project" are available, they should not since there is no namespace nor project selected.

adietish:
fixes #812

### DIFF
--- a/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/actions/SetAsCurrentClusterAction.kt
+++ b/src/main/kotlin/com/redhat/devtools/intellij/kubernetes/actions/SetAsCurrentClusterAction.kt
@@ -19,7 +19,7 @@ import com.redhat.devtools.intellij.kubernetes.telemetry.TelemetryService
 import com.redhat.devtools.intellij.kubernetes.telemetry.TelemetryService.NAME_PREFIX_CONTEXT
 import javax.swing.tree.TreePath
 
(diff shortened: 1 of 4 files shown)


sbouchet (commented):
LGTM. 
nitpicking is when the selected context is the one currently used, the "Set current namespace/project" should not be enabled. to be tackle in another issue.

sbouchet (approved):

sbouchet (changes requested):
second thought, there is something wrong with this, i can see both set current namespace and set current project on the current namespace root node ...

adietish:
> i can see both set current namespace and set current project on the current namespace root node

@sbouchet: good catch, thx! The visibility check in `SetCurrentProject` was using the enabled check :(

```kotlin
    override fun isVisible(selected: Array<out Any>?): Boolean {
        return !selected.isNullOrEmpty()
==>             && super.isEnabled(selected)
    }
```

Both your complaints were fixed when correcting this. Please re-review.

sbouchet (approved):
No more extra actions in empty selection with this fix. LGTM

PR #827, opened 2025-01-31
fix: copy KUBECONFIG as defined in shell to system properties (#826)

### LINKED ISSUE #826: MacOS: when launched within IDE (not run from source) KUBECONFIG env variable is not available
**Steps:**
1. ASSERT: use `MacOS`, only MacOS and maybe `Windows` are affected
2. ASSERT: have `KUBECONFIG` env var set in `~/.zshrc` or `./zprofile` 
Setup is documented here:
* https://github.com/redhat-developer/intellij-openshift-connector/issues/966#issue-2819190589 or
* https://github.com/redhat-developer/intellij-kubernetes/issues/779#issuecomment-2265980362 
3. ASSERT: have latest plugin installed in your IDE (**binary, not launched from source**)
4. EXEC: open Kubernetes tool window 

**Result:**
No contexts are displayed, the window stays empty even though `KUBECONFIG` is set ex. in the IDEA terminal (`echo $KUBECONFIG` returns your list of kube config files).

Image

adietish:
fixes #826


fbricon (approved):
LGTM

PR #841, opened 2025-02-24
fix: report refreshing all context instead of null (#839)

### LINKED ISSUE #839: Refreshing all context causes notification message "Refresh null"
**Steps:**
1. ASSERT: have all your contexts listed
2. EXEC: right click outside of the listed context and pick "Refresh" in the ctx menu

**Result:**
The list of all contexts is refreshed. A message is displayed in the status bar "Refreshing null". If you click on it, a notification is displayed with the same message:

Image

adietish:
fixes #839


sbouchet (approved):
LGTM

PR #850, opened 2025-02-26
chore: remove unneeded IC-2022.3 compat (minVersion now: 2023.1)

adietish:
removes a reflection based bridge that was needed when we still supported IC-2022.3. We're now at >= IC-2023.1


sbouchet (approved):
LGTM

PR #874, opened 2025-05-13
chore(skip-release): bump intellij-common to 1.9.9

adietish:


adietish:
intellij-common 1.9.9 is causing exception in jacks with editors. This was caused by a bump in jackson 2.19 for intellij-common. We therefore reverted this bump: https://github.com/redhat-developer/intellij-common/pull/292
Closing this PR.

PR #887, opened 2025-05-21
chore: create dependabot workflow

### LINKED ISSUE #804: Create dependabot workflow
similar to https://github.com/redhat-developer/intellij-openshift-connector/blob/main/.github/dependabot.yml and https://github.com/redhat-developer/intellij-openshift-connector/blob/main/.github/workflows/dependency-submission.yml

sbouchet:
fixes #804


adietish (approved):
Awesomeness! This will automate things furthermore, less work 😃

